PR #333, opened 2017-03-15
Welcome email template, closes #329

### LINKED ISSUE #329: create MediaCloud email template
We need a quick-and-dirty email template so our emails don't look terrible anymore.  @kma2 I need to pull you to work on this very quickly - shouldn't take long.  I just shared via Google Docs an example of the type of content we need to support.  Feel free to connect with Natalie on this directly.  

I've found [MailChimp's email table layout notes](https://templates.mailchimp.com/development/html/) super helpful in the past.  [`static/img`](https://github.com/mitmedialab/MediaCloud-TopicMapper/tree/master/server/static/img) has some logos.

It you make a quick template with header, body, and footer content areas then we can slap in email text as needed.  The output should be an HTML page with placeholder spots for title and content that our back-end team can use in their code (and we can too).  Commit it into `server/templates/email` for now.

kma2:


kma2:
email_template

kma2:
email_template

kma2:
The HTML commented-out version is there in case we want to easily modify the CSS in the <style> tags later, without having to manually edit each inline CSS rule. We can modify the currently commented out code, copy/paste it into MailChimp's Inliner tool (https://templates.mailchimp.com/resources/inline-css), and convert everything back to inline CSS.

Issue #530, opened 2017-07-14
don't let users create topic if name already used


rahulbot:
There is a DB-level rule that requires unique topic names.  We need to make sure that the name isn't used already before someone creates a topic.  So in the topic creation process, on the preview screen we probably need to have the server check if the name is already in use and fore the user to change the name before proceeding.  It should show them an error saying something like "Sorry, that topic name is already in use.  Please choose another"

cindyloo:
in TopicCreate1ConfigureContainer, I could dispatch a topic name search before we allow the user to proceed to the next step, or we could dispatch it when the name field in the form loses focus.. I suggest this because it seems we could and shouldmore quickly inform the user the name is already taken, also because the preview page currently doesn't have the name field on it, and is gigantic to begin with

cindyloo:
https://github.com/erikras/redux-form/issues/2799

cindyloo:
another helpful article:
[http://jamesknelson.com/grokking-es6-promises-the-four-functions-you-need-to-avoid-callback-hell/](url)
The fact that rejection handlers return a resolved promise by default caused me a lot of pain when I was first learning – don’t let it happen to you. 
The take-away here is that if you want to process your errors, make sure you don’t just return the processed value: reject the processed value. So, instead of:

`return 'some description of :('`
Use this magic incarnation to return a failed promise with a given value instead of a successful one:

// I'll explain this a little later
`return Promise.reject({anything: 'anything'});`

PR #534, opened 2017-07-18
Fix warn duplicate topic names #530

cindyloo:


cindyloo:
fixed and ready for #530

PR #790, opened 2018-01-22
Feature explorer save searches

cindyloo:
Need to check nickname for deleted queries, etc…

### DIFF
--- a/server/views/explorer/explorer_query.py
+++ b/server/views/explorer/explorer_query.py
@@ -74,14 +74,21 @@ def api_explorer_demo_collections_by_ids():
     return jsonify(coll_list)
 
 
-@app.route('/api/explorer/saveQuery', methods=['GET'])
+@app.route('/api/explorer/saveSearches', methods=['GET'])
 @flask_login.login_required
 @arguments_required('queryName', 'timestamp', 'queryParams')
-def save_user_query():
+def save_user_search():
     username = user_name()
-    # TODO any checking here?
-    db.add_item_to_users_list(username, 'queries', request.args)
-    return jsonify({'savedQuery': request.args['query_string']})
+    db.add_item_to_users_list(username, 'searches', request.args)
+    return jsonify({'savedQuery': request.args['queryName']})
+
+@app.route('/api/explorer/loadUserSearches', methods=['GET'])
+@flask_login.login_required
+def load_user_searches():
+    username = user_name()
+    search_list = db.get_users_lists(username, 'searches')
+    return jsonify(search_list)
+
 
 
 # TODO use this or the other collection list retrieval?
--- a/server/views/explorer/explorer_query.py
+++ b/server/views/explorer/explorer_query.py
@@ -76,7 +76,7 @@ def api_explorer_demo_collections_by_ids():
 
 @app.route('/api/explorer/saveQuery', methods=['GET'])
 @flask_login.login_required
-@arguments_required('label', 'query_string')
+@arguments_required('queryName', 'timestamp', 'queryParams')
 def save_user_query():
     username = user_name()
     # TODO any checking here?
--- a/server/views/explorer/explorer_query.py
+++ b/server/views/explorer/explorer_query.py
@@ -89,6 +89,13 @@ def load_user_searches():
     search_list = db.get_users_lists(username, 'searches')
     return jsonify(search_list)
 
+@app.route('/api/explorer/deleteSearch', methods=['GET'])
+@flask_login.login_required
+@arguments_required('queryName', 'timestamp')
+def delete_user_search():
+    username = user_name()
+    result = db.remove_item_from_users_list(username, 'searches', request.args['queryName'])
+    return jsonify(result)
 
 
 # TODO use this or the other collection list retrieval?
--- a/server/views/explorer/explorer_query.py
+++ b/server/views/explorer/explorer_query.py
@@ -91,10 +91,10 @@ def load_user_searches():
 
 @app.route('/api/explorer/deleteSearch', methods=['GET'])
 @flask_login.login_required
-@arguments_required('queryName', 'timestamp')
+@arguments_required('queryName', 'timestamp', 'queryParams')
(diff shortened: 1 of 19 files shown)


server/views/explorer/explorer_query.py
```diff
@@ -74,14 +74,28 @@ def api_explorer_demo_collections_by_ids():
...
+@arguments_required('queryName', 'timestamp', 'queryParams')
+def save_user_search():
     username = user_name()
-    # TODO any checking here?
-    db.add_item_to_users_list(username, 'queries', request.args)
-    return jsonify({'savedQuery': request.args['query_string']})
+    db.add_item_to_users_list(username, 'searches', request.args)
+    return jsonify({'savedQuery': request.args['queryName']})
+
+@app.route('/api/explorer/loadUserSearches', methods=['GET'])
```
rahulbot: Most of our code uses hyphens on URL paths (ie. `load-user-searches`).  Lets standardize on that.

server/views/explorer/explorer_query.py
```diff
@@ -74,14 +74,28 @@ def api_explorer_demo_collections_by_ids():
...
+@flask_login.login_required
+def load_user_searches():
+    username = user_name()
+    search_list = db.get_users_lists(username, 'searches')
+    return jsonify(search_list)
+
+@app.route('/api/explorer/deleteSearch', methods=['GET'])
```
rahulbot: Same comment here - lets use `delete-search` as the url instead

src/components/explorer/builder/QueryPicker.js
```diff
@@ -155,9 +156,24 @@ class QueryPicker extends React.Component {
...
+  saveThisSearch = (queryName) => {
+    const { queries, sendAndSaveUserSearch } = this.props; // formQuery same as selected
+    // filter out removed ids...
+    const searchstr = generateQueryParamString(queries.map(q => ({
+      label: q.label,
+      q: q.q,
+      color: q.color,
+      startDate: q.startDate,
+      endDate: q.endDate,
+      sources: q.sources, // de-aggregate media bucket into sources and collections
```
rahulbot: this comment seems irrelevant

src/components/explorer/builder/QueryPicker.js
```diff
@@ -255,6 +271,8 @@ class QueryPicker extends React.Component {
           <QueryForm
             initialValues={selected}
             selected={selected}
+            searchNickname={queries.map(q => q.label).join(',')}
```
rahulbot: Do you think `', '` would be more readable (note the space after the comma)

server/views/explorer/explorer_query.py
```diff
@@ -74,14 +74,28 @@ def api_explorer_demo_collections_by_ids():
...
+@flask_login.login_required
+@arguments_required('queryName', 'timestamp', 'queryParams')
+def delete_user_search():
+    username = user_name()
+    result = db.remove_item_from_users_list(username, 'searches', request.args)
+    return jsonify(result.raw_result)
```
rahulbot: This should probably be an object like `{ 'success': results.raw_result }`, because I can't think of many places where we return just a single value in json to the browser.

src/components/explorer/builder/QueryPicker.js
```diff
@@ -342,19 +366,30 @@ const mapDispatchToProps = dispatch => ({
...
+  handleLoadSelectedSearch: (selectedSearch) => {
+    if (selectedSearch && selectedSearch.queryParams) {
+      dispatch(push({ pathname: '/queries/search', search: `?q=${selectedSearch.queryParams}` }));
+    }
+  },
+  handleDeleteUserSearch: (selectedSearch) => {
+    if (selectedSearch && selectedSearch.queryName) {
+      dispatch(deleteUserSearch(selectedSearch))
+      .then((results) => {
+        if (results) dispatch(loadUserSearches());
```
rahulbot: following my previous comment, this would then be `if (results.success)`.  And it should have an `else` that dispatches an `updateFeedback` action saying it couldn't delete the search for some reason.

src/components/explorer/builder/QueryPickerCustomQueryHandler.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { injectIntl } from 'react-intl';
+import QueryPickerLoadUserSearchesDialog from './QueryPickerLoadUserSearchesDialog';
+import QueryPickerSaveUserSearchesDialog from './QueryPickerSaveUserSearchesDialog';
+
+
+const QueryPickerCustomQueryHandler = (props) => {
```
rahulbot: This name doesn't make sense to me.  I think this is the wrapper that holds the load and delete saved searches buttons.  And it doesn't have any state, so I don't think it should include `Container`. Is a more descriptive name `SavedSearchControls`?

src/reducers/explorer/savedSearches.js
```diff
@@ -0,0 +1,14 @@
+import { createAsyncReducer } from '../../lib/reduxHelpers';
+import { LOAD_USER_SEARCHES } from '../../actions/explorerActions';
+
+const searches = createAsyncReducer({
+  initialState: {
+    list: [],
+  },
+  action: LOAD_USER_SEARCHES,
+  handleSuccess: payload => ({
+    list: payload,
```
rahulbot: Why does this have a `handleSuccess`?  Shouldn't the server just return `{list: allSearches}`? Lets try to avoid custom handlers here unless they do some business logic (ie. change the data). It just makes debugging on the server/browser harder.

cindyloo:
so - if the user loads and selects a search, because it is a push, we don't keep the search "nickname." if the user modifies this search and then saves it, we will create a new saved search. The workaround is that the user removes older queries. Not a big issue but otherwise we'd want to push the saved search nickname into the url...

cindyloo:
- we will allow a dumb save for now- leaving it up to the user to manage saved searches by name

PR #795, opened 2018-01-24
Feature topic max stories

cindyloo:


src/components/topic/create/TopicDetailForm.js
```diff
@@ -133,6 +134,22 @@ const TopicDetailForm = (props) => {
           {queryWarning}
         </Col>
       </Row>
+      <Row>
+        <Col lg={3}>
+          <Permissioned onlyRole={PERMISSION_TOPIC_ADMIN}>
```
rahulbot: `PERMISSION_TOPIC_ADMIN` means the person is an admin for this topic.  I think we want `PERMISSION_ADMIN` here - ie. any system admin can change this limit while creating a topic.  People shouldn't be able to change this themselves.

cindyloo:
merging with EditTopicContainer updates related to MediaPicker integration...

PR #799, opened 2018-01-25
Feature advanced search media picker

cindyloo:
Incorporating max_stories (#795 as well)
One last item: handling the need to refresh topicInfo between topicSummaryContainer and EditTopicContainer


server/views/topics/apicache.py
```diff
@@ -131,7 +131,6 @@ def topic_word_counts(user_mc_key, topics_id, **kwargs):
         'foci_id': foci_id,
         'q': q,
         'sample_size': DEFAULT_WORD_COUNT_SAMPLE_SIZE,
-        'num_words': 500
```
rahulbot: Why are you removing this default setting?

server/views/topics/maps.py
```diff
@@ -34,8 +34,8 @@ def map_files(topics_id):
         rendered_filename = prefix+".gexf"
         # check if rendered file is there
         is_rendered = os.path.isfile(os.path.join(DATA_DIR,rendered_filename))
-        # logger.warn(os.path.join(DATA_DIR,rendered_filename))
-        # logger.warn(is_rendered)
+        logger.warn(os.path.join(DATA_DIR,rendered_filename))
+        logger.warn(is_rendered)
```
rahulbot: These should be commented out, or switched to debug logs, no?

src/components/topic/create/EditTopicContainer.js
```diff
@@ -110,8 +188,7 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
             dispatch(setTopicNeedsNewSnapshot());
           }
           // update topic info and redirect back to topic summary
-          dispatch(fetchTopicSummary(results.topics_id))
```
rahulbot: I'm surprised that here you don't need to update the topic summary anymore?  Do the changes you just made show up on the /summary page after saving without doing this call?

src/components/topic/create/TopicDetailForm.js
```diff
@@ -133,6 +134,22 @@ const TopicDetailForm = (props) => {
...
       </Row>
+      <Row>
+        <Col lg={3}>
+          <Permissioned onlyRole={PERMISSION_ADMIN}>
+            <Field
+              name="max_stories"
+              component={renderTextField}
+              type="inline"
+              fullWidth
+              value="1000000"
```
rahulbot: I think you have one extra zero here - we want the default value here to be 100,000, not 1,000,000

src/components/topic/create/TopicForm.js
```diff
@@ -128,9 +128,12 @@ function validate(values, props) {
 const asyncValidate = (values, dispatch) => (
   dispatch(fetchTopicSearchResults(values.name))
     .then((results) => {
+      const checkForNameMatches = results.topics.filter(t => t.name.toLowerCase() === values.name.toLowerCase());
+      const checkForIdMatches = results.topics.filter(t => t.id !== values.topics_id);
```
rahulbot: I think we want to check carefully here to make sure the name that matches isn't the same as the id that matches.  Maybe it doesn't matter from a Boolean logic point of view, but from a human logic one I think it does.  To be concrete, I think we want this combined check:
```javascript
const otherTopicsWithSameName =
  results.topics.filter(t => (t.name.toLowerCase() === values.name.toLowerCase())
    && (t => t.id !== values.topics_id));
```

src/reducers/topics/selected/selected.js
```diff
@@ -28,7 +28,7 @@ function id(state = null, action) {
 function needsNewSnapshot(state = false, action) {
   switch (action.type) {
     case SET_TOPIC_NEEDS_NEW_SNAPSHOT:
-      return action.payload;
+      return action.payload ? action.payload : {};
```
rahulbot: I'm confused about this.  Where we use this value it expects a boolean.  This needs to store a boolean to the store indicating if this needs a new snapshot or not.

src/components/topic/create/EditTopicContainer.js
```diff
@@ -110,8 +188,7 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
             dispatch(setTopicNeedsNewSnapshot());
```
rahulbot: Doesn't this need to pass a value  of `true` to `setTopicNeedsNewSnapshot`? Or am I confused about how that action works?

src/components/topic/create/EditTopicContainer.js
cindyloo: this is one of the issues I mentioned- without this call, the page renders but with old data. If we call the fetch before the push, we get an empty topic summary page. What's the right way to handle refreshing of the topicInfo?

src/components/topic/create/EditTopicContainer.js
cindyloo: after reviewing the PropTypes and the reducer, we do want to send 'true' as a parameter otherwise it returns null and causes a crash

src/components/topic/create/TopicForm.js
cindyloo: nice!

src/reducers/topics/selected/info.js
```diff
@@ -3,9 +3,9 @@ import { createAsyncReducer } from '../../../lib/reduxHelpers';
 
 const info = createAsyncReducer({
   action: FETCH_TOPIC_SUMMARY,
-  SET_TOPIC_FAVORITE_RESOLVED: (payload, state) => ({
+  UPDATE_TOPIC_RESOLVED: payload => ({ ...payload }),
+  SET_TOPIC_FAVORITE_RESOLVED: payload => ({
```
rahulbot: Did you verify that favoriting topics still works after this change? Just to be sure...

cindyloo:
the info.js selector seems to be working properly  - it properly toggles favorited from any in the topic list. However, there is an error in the all.js selector - which is activated when the star in the _subheader_ is clicked..

Issue #807, opened 2018-01-29
create subtopic technique by media type


rahulbot:
We should have new subtopic technique the lets you create subtopics by media type. This should look exactly like the partisan-retweet one, but use the media type metadata.

*(note from recent Berkman mtg)*

cindyloo:
I was making the assumption that they would want to be able to select media type regarding the pie-chart story_count/total_stories, but that probably isn't necessary, right?

rahulbot:
No - no selection on this. Works like the "retweet partisanship" one, which is just creates 5 subtopics automatically (one for each tag in`TAG_SETS_ID_MEDIA_TYPE = 1972`)

Issue #804, opened 2018-01-29
remove story bitly counts


rahulbot:
We don't want to show these anymore in Explorer or Topic Mapper UIs, because they are neither particularly representative nor particularly useful anymore.  This should just involve changing the relevant `StoryTable` Components.

*(note from recent Berkman mtg)*

rahulbot:
Note: leave them in the CSV download, just in case someone wants them.  If we decide to remove them completely from Media Cloud then we'll remove them from the CSV.  That conversation is happening on https://github.com/berkmancenter/mediacloud/issues/274

Issue #803, opened 2018-01-29
add tweet counts to Topic story table


rahulbot:
If a topic has `ch_monitor_id` set, then on the `StoryTable` add a column for "Tweet Counts" and fill it with the value in `story.simple_tweet_count`.

Also add this `simple_tweet_count` under a new "tweet count" header to every story list download (no matter if `ch_monitor_id` is set or not.

*(note from recent Berkman mtg)*

cindyloo:
in almost all of these invocations, TopicStoryTable is several layers deep and doesn't have access to the topic values unless we push it from each container all the way down. Can we just evaluate story.simple_tweet_count whether is exists or not?

cindyloo:
eg this commit.

rahulbot:
I imagine this component is always a child in a container (which has state). How about adding a ‘showTweetCount’ to this component. Then in the containers pull in the ‘ch_monitor_id’ and use that to decide the value for ‘showTweetCount’. What do you think?

rahulbot:
NO, inferring this from the data is gonna break because there are stories from twitter-connected topics that also show up in non-twitter-connected topics. We have to decide what to show based on the topic config (sorry).

cindyloo:
ok I thought that could be the case
so in `mapStateToProps` for the 9 containers eg MediaInlink, -Outlink,etc, do something like
showTweetCount: state.topics.selected.info.ch_monitor_id.
and pass that to TopicStoryCount...

rahulbot:
Or if this is every use of the component that needs to be wrapped, why not add a container around it that loads that one piece from the store and passes on all the props to the component. Then that container would be what you use everywhere, without changing props sent into it. Does that suggestion make sense?

cindyloo:
that makes sense as well. I did however just update those files but I need a topic that has tweets in order to properly test it and have looked through several already. Do you have one you typically use - I haven't set up a crimson hexagon account..

cindyloo:
completed

PR #816, opened 2018-02-01
Feature topic bugs from berkman

cindyloo:
#803, #804, #807

### DIFF
--- a/server/views/topics/foci/mediatype.py
+++ b/server/views/topics/foci/mediatype.py
@@ -21,18 +21,23 @@ def media_type_story_counts(topics_id):
     media_type_tags = cached_media_tags(TAG_SETS_ID_MEDIA_TYPE)
     # grab the total stories
     total_stories = topic_story_count(user_mediacloud_key(), topics_id)['count']
-    # make a count for each tag based on media_od
+    # make a count for each tag based on media_id
     for tag in media_type_tags:
-        tagged_story_count = topic_story_count(user_mediacloud_key(), topics_id, q=tag['media_query'])['count']
+        
+        tag_media_ids = " ".join(tag['media_ids'])
+        media_ids_query_clause = "media_id:({})".format(tag_media_ids)
+        if len(tag['media_ids']) > 0:
+            tagged_story_count = topic_story_count(user_mediacloud_key(), topics_id, q=media_ids_query_clause)['count']
+        else:
+            tagged_story_count = 0
         tag_story_counts.append({
             'label': tag['label'],
             'tags_id': tag['tags_id'],
             'count': tagged_story_count,
             'pct': float(tagged_story_count)/float(total_stories)
         })
-    # order them in the way a person would expect ( left to center to right)
-    ordered_tag_story_counts = list()
-    return jsonify({'story_counts': ordered_tag_story_counts})
+
+    return jsonify({'story_counts': tag_story_counts})
 
 
 @app.route('/api/topics/<topics_id>/focal-sets/media-type/preview/coverage', methods=['GET'])
--- a/server/views/topics/foci/mediatype.py
+++ b/server/views/topics/foci/mediatype.py
@@ -1,0 +1,98 @@
+import logging
+from flask import jsonify, request
+import flask_login
+
+from server import app
+from server.util.request import api_error_handler, json_error_response, form_fields_required, arguments_required
+from server.views.topics.apicache import topic_story_count
+from server.auth import user_mediacloud_key, user_mediacloud_client
+from server.views.topics.apicache import topic_tag_coverage, _cached_topic_tag_counts, cached_topic_timespan_list
+from server.views.topics.foci import FOCAL_TECHNIQUE_BOOLEAN_QUERY
+from server.util.tags import TAG_SETS_ID_MEDIA_TYPE
+import json
+logger = logging.getLogger(__name__)
+
+DEFAULT_SAMPLE_SIZE = 5000
+
+def get_media_type_by_sentence_field_counts(topics_id, num_countries):
+    tag_media_type_counts= []
+
+    # get the total stories for a topic
+    total_stories = topic_story_count(user_mediacloud_key(), topics_id)['count']
+
+    # get the top countries by the sentence field counts iwth overall timespan
+    timespans = cached_topic_timespan_list(user_mediacloud_key(), topics_id)
+    overall_timespan = [t for t in timespans if t['period'] == "overall"]
+    overall_timespan = next(iter(overall_timespan))
+    timespan_query = "timespans_id:{}".format(overall_timespan['timespans_id'])
+    top_media_type_tags = _cached_topic_tag_counts(user_mediacloud_key(), topics_id, TAG_SETS_ID_MEDIA_TYPE, DEFAULT_SAMPLE_SIZE, timespan_query)
+    
+    # get the total stories for a topic
+    total_stories = topic_story_count(user_mediacloud_key(), topics_id)['count']
+
+    top_media_type_tags = top_media_type_tags[:num_themes]
+    # for each country, set up the requisite info for UI
+    for tag in top_media_type_tags:
+        tag_media_type_counts.append({
+            'label': tag['label'],
+            'tags_id': tag['tags_id'],
+            'count': tag['count'],
+            'pct': float(tag['count']) / float(total_stories), #sentence_field_count / total story per topic count
+        })
+
+    return tag_media_type_counts
+
+
+@app.route('/api/topics/<topics_id>/focal-sets/media-type/preview/story-counts', methods=['GET'])
+@flask_login.login_required
+@arguments_required('numCountries')
+@api_error_handler
+def media_type_story_counts(topics_id):
+    # using sentence field count to approximate story mentions by top country
+    num_media_type = int(request.args['numMediaType'])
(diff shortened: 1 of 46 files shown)


src/components/topic/snapshots/foci/CreateFocusContainer.js
```diff
@@ -20,12 +20,13 @@ const localMessages = {
...
   topCountriesFocusSaved: { id: 'focus.create.booleanSaved', defaultMessage: 'We created a new subtopics by top countries.' },
   nytFocusSaved: { id: 'focus.create.booleanSaved', defaultMessage: 'We created a new subtopics with NYT Themes tags.' },
+  mediaTypeFocusSaved: { id: 'focus.create.mediaType.saved', defaultMessage: 'We created a new subtopics with Media Type tags.' },
 };
 
 const CreateFocusContainer = (props) => {
   const { topicId, location, handleDone } = props;
   const { focalSetDefId, focalTechnique } = props.location.query;
-  const initialValues = { numberSelected: DEFAULT_SELECTED_NUMBER };
+  const initialValues = { numberSelected: DEFAULT_SELECTED_NUMBER, mediaType: 0 };
```
rahulbot: what is the new `mediaType` value for?

src/components/topic/snapshots/foci/CreateFocusContainer.js
```diff
@@ -116,6 +118,16 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
             dispatch(push(`/topics/${topicId}/snapshot/foci`)); // go back to focus management page
             dispatch(reset('snapshotFocus')); // it is a wizard so we have to do this by hand
           });
+      case FOCAL_TECHNIQUE_MEDIA_TYPE: // TODO, not yet tested
+        saveData = { ...formValues, data: queryData.mediaType.map(c => ({ tags_id: c.tags_id, label: c.label })) };
```
rahulbot: Why is this sending tag ids to the server? I think all we need to know is that they want a media-type focal set, right?

src/components/topic/snapshots/foci/builder/mediaType/EditMediaTypeContainer.js
```diff
@@ -0,0 +1,106 @@
...
+const localMessages = {
+  title: { id: 'focus.create.edit.title', defaultMessage: 'Step 2: Preview Subtopics by Media type' },
+  about: { id: 'focus.create.edit.about',
+    defaultMessage: 'This will create a set of subtopics as filtered by media types.' },
+};
+
+const EditMediaTypeContainer = (props) => {
+  const { topicId, onPreviousStep, handleSubmit, finishStep, formData, initialValues } = props;
+  const { formatMessage } = props.intl;
+  let mediaTypeSelected = initialValues.mediaType;
```
rahulbot: I think we need to remove this `mediaTypeSelected` thing... you're picking the whole set, not one of the tags in the set.

src/components/topic/snapshots/foci/builder/mediaType/MediaTypeCoveragePreviewContainer.js
```diff
@@ -0,0 +1,97 @@
...
+import composeAsyncContainer from '../../../../../common/AsyncContainer';
+import { fetchCreateFocusMediaTypeCoverage } from '../../../../../../actions/topicActions';
+import DataCard from '../../../../../common/DataCard';
+import PieChart from '../../../../../vis/PieChart';
+import { getBrandDarkColor } from '../../../../../../styles/colors';
+
+const localMessages = {
+  title: { id: 'topic.snapshot.mediaType.coverage.title', defaultMessage: 'Story Coverage' },
+  intro: { id: 'topic.snapshot.mediaType.coverage.intro', defaultMessage: 'By Media Type' },
+  included: { id: 'topic.snapshot.mediaType.coverage.matching', defaultMessage: 'Stories about these media types for this topic' },
```
rahulbot: Change this to "Stories from a media source that has a Media Type"

src/components/topic/snapshots/foci/builder/mediaType/MediaTypeCoveragePreviewContainer.js
```diff
@@ -0,0 +1,97 @@
...
+  notIncluded: { id: 'topic.snapshot.mediaType.coverage.total', defaultMessage: 'All Stories about this topic' },
```
rahulbot: Change this to "Stories from a media source that has NO Media Type set"

src/components/topic/snapshots/foci/builder/mediaType/MediaTypeCoveragePreviewContainer.js
```diff
@@ -0,0 +1,97 @@
...
+    );
+  }
+}
+
+MediaTypeCoveragePreviewContainer.propTypes = {
+  // from compositional chain
+  intl: PropTypes.object.isRequired,
+  // from parent
+  topicId: PropTypes.number.isRequired,
+  mediaTypeSelected: PropTypes.number.isRequired,
```
rahulbot: I think this is to be removed.

src/components/topic/snapshots/foci/builder/mediaType/MediaTypeStoryCountsPreviewContainer.js
```diff
@@ -0,0 +1,95 @@
...
+    const { counts } = this.props;
+    const { formatNumber } = this.props.intl;
+    let content = null;
+    if (counts !== null) {
+      const data = counts.map((info, idx) => ({
+        value: info.count,
+        fill: mapD3Top10Colors(idx),
+        rolloverText: `${info.label}: ${formatNumber(info.count)}`,
+      }));
+      content = (<PackedBubbleChart
```
rahulbot: Since there are only 5 types, I think we want a `BubbleRowChart`, not a packed one.

server/views/topics/foci/mediatype.py
```diff
@@ -0,0 +1,92 @@
...
+    focal_set_name = request.form['focalSetName']
+    focal_set_description = request.form['focalSetDescription']
+    country_data = json.loads(request.form['data[]'])
+    focal_technique = FOCAL_TECHNIQUE_BOOLEAN_QUERY
+    new_focal_set = user_mc.topicFocalSetDefinitionCreate(topics_id, focal_set_name, focal_set_description, focal_technique)
+    if 'focal_set_definitions_id' not in new_focal_set:
+        return json_error_response('Unable to create the subtopic set')
+    # now make the foci in it - one for each country
+    focus_def_results = []
+    for tag in country_data:
```
rahulbot: I think we want to iterate over the tags in the MEDIA_TYPE tag set here, not over data from the client.

server/views/topics/foci/mediatype.py
```diff
@@ -0,0 +1,92 @@
...
+        params = {
+            'name': tag['label'],
+            'description': u"Stories about {}".format(tag['label']),
```
rahulbot: I think this should say `u"Stories from {} sources".format(tag['label'])`

server/views/topics/foci/mediatype.py
```diff
@@ -0,0 +1,92 @@
...
+            'query': "tags_id_stories:{}".format(tag['tags_id']),
```
rahulbot: I think this needs to be a query for `tags_id_media:{}`, no? Because we want to find stories that are from *media*  that have the media type tag (i.e. "broadcast", etc).

src/components/topic/snapshots/foci/CreateFocusContainer.js
cindyloo: right - removed all related code/params because we are not specifying individual media types

cindyloo: it makes more sense to use the mediaType ids that are cached on the server versus passing the tags_id back. We do have to pass the new user-defined label back to the server, but it is the same for all media types.

cindyloo:
ok updated code as suggested

cindyloo:
just a note: the query for media types is tags_id_media whereas some of the other focal sets queries are tags_id_stories...

server/views/topics/foci/mediatype.py
```diff
@@ -47,14 +47,13 @@ def media_type_coverage(topics_id):
     media_type_tags = cached_media_tags(TAG_SETS_ID_MEDIA_TYPE)
```
rahulbot: This is still calling `cached_media_tags`... we want to remove that slow call, because we don't need the list of media sources here.  We can use the tag ids for everything.

server/views/topics/foci/mediatype.py
```diff
@@ -66,25 +65,25 @@ def cached_media_tags(tag_sets_id):
...
-@form_fields_required('focalSetName', 'focalSetDescription', 'data[]')
+@form_fields_required('focalSetName', 'focalSetDescription')
 @flask_login.login_required
 def create_media_type_focal_set(topics_id):
     user_mc = user_mediacloud_client()
     # grab the focalSetName and focalSetDescription and then make one
     focal_set_name = request.form['focalSetName']
     focal_set_description = request.form['focalSetDescription']
-    country_data = json.loads(request.form['data[]'])
+    media_type_tags = cached_media_tags(TAG_SETS_ID_MEDIA_TYPE)
```
rahulbot: same here - we need a list of the tags, not the media in the tags

cindyloo:
update to mediatype.py pushed

Issue #868, opened 2018-02-27
comparative word cloud doesn't update on new query run


rahulbot:
probably missing a newProps handler

PR #928, opened 2018-03-09
fix for #868

cindyloo:
If from the search button, act as if a new page mount,
If from the UI comparison widget, switch out the word clouds


cindyloo:
better test in componentWillMount, which does get invoked more than once.

Issue #943, opened 2018-03-12
Query is reset to "*" after clicking on other controls


pypt:
Hi guys,

Clicking around Explorer, encountered an issue where query gets reset to `*` after entering it and then touching some other control afterwards.

To reproduce:

1. Enter a query, e.g. "racism"
2. Edit any of the dates in the "For dates"
3. Query gets reset to "*"

or

1. Enter a query, e.g. "racism"
2. Choose collection, e.g. "ABYZ Global English Sources"
3. Remove the default collection ("US Top News")
4. Query is reset to "*"

cindyloo:
nice find, Linas. I needed to include the updated but unstored edits to the queryForm like this:
```
const updatedQuery = {
      ...query,
      ..formQuery,
}
```
in the handling of updates to the various form fields.Due to the complexity of the media handling and dates, we have different methods for each field, but this addresses them all similarly

rahulbot:
Connie also mentioned this one.

PR #945, opened 2018-03-12
#943 update with form data included

cindyloo:


cindyloo:
this is ready to go @rahulbot

Issue #974, opened 2018-03-20
add stories per week to media source detail page


rahulbot:
After the "Health" box is removed (in #913), lets replace it with a "Stories per Week" box.  I think this number is available in the health data we have already.  Make sure to `Math.round` it so we don't look dumb by talking about fractional stories ;-)

PR #975, opened 2018-03-20
Source bug fixes

cindyloo:


cindyloo:
also #974

src/components/source/collection/CollectionDetailsContainer.js
```diff
@@ -35,8 +37,10 @@ class CollectionDetailsContainer extends React.Component {
 
   searchOnDashboard = () => {
```
rahulbot: I'd suggest renaming this `searchOnExplorer` so we don't confuse ourselves

src/lib/dateUtil.js
```diff
@@ -48,6 +48,11 @@ export function oneWeekLater(date) {
   return weekLater;
 }
 
+export function oneMonthBefore(date) {
+  const monthBefore = moment(date).subtract(1, 'month').format(TOPIC_DATE_FORMAT);
```
rahulbot: I wouldn't expect this function to format the date for me, but rather to return a date or moment object of the caller to format as it sees fit.

src/components/source/mediaSource/SourceSentenceCountContainer.js
```diff
@@ -51,7 +50,6 @@ class SourceSentenceCountContainer extends React.Component {
           health={health}
           height={250}
           filename={filename}
-          lineColor={getBrandDarkColor()}
```
rahulbot: Why are you removing the line color of this chart? I think the change you made to the `.health-plot-line` CSS class is correct.  If you're going to remove this here, then it would make sense to edit `AttentionOverTimeChart` to default its line color internally to `getBrandDarkColor` if none is specified.  That actually makes sense to me, since a number of our other visualizations work that way.

src/lib/dateUtil.js
cindyloo: the routines here return a smattering of formatted and unformatted dates so we determined this is ok

src/components/source/mediaSource/SourceSentenceCountContainer.js
cindyloo: updated the AttentionOverTimeChart lineColor to handle default line color and removed the default passed in from the containers

PR #983, opened 2018-03-22
Source bug fixes

cindyloo:


cindyloo:
what should the exploroer query label be coming from topics/sources/collections?

PR #1025, opened 2018-04-10
Fix topic edit and mediapicker selection

cindyloo:
Only remaining issue is that the pristine flag is not set to false when the sub-form sourcesAndCollections is modified


cindyloo:
note that EditTopicContainer at lines 206-211 invokes the resetTopic call. This is not implemented in master. It fails safely, but perhaps we should comment it out for now? lmk

cindyloo:
the user would have to modify something (anything) else that isn't a field array field. so, it's a small bug but they can edit and save a topic

PR #1146, opened 2018-06-07
Feature word option menu sample size

cindyloo:


cindyloo:
re-architected!

cindyloo:
completed, updated filter handling (q, sample_size and filters integration)

PR #1165, opened 2018-06-18
Rename "feed_status" to "active" boolean, "feed_type" to "type"

pypt:
Hi again Cindy & Rahul,

As part of our cleanup (https://github.com/berkmancenter/mediacloud/issues/416), we want to:

1. Rename `feed_status` API JSON response field to `active` boolean, and
2. Rename `feed_type` API JSON response field to `type`.

*Before:*

```json
{
    "feed_status": "active",
    "feed_type": "syndicated"
}
```

*After:*

```json
{
    "active": true,
    "type": "syndicated"
}
```

(Admittedly I didn't even test those changes by running them because I'm not much of a frontender, those commits are basically just a careful find-and-replace, so I would be thankful if you could have a hard look into them.)

As detailed in https://github.com/berkmancenter/mediacloud/pull/421, could you:

1. Wait for us to release said PR (https://github.com/berkmancenter/mediacloud/pull/421),
2. Review, merge & deploy this PR on your end,
3. Let us know when you did so?

### DIFF
--- a/server/views/sources/collection.py
+++ b/server/views/sources/collection.py
@@ -146,7 +146,7 @@ def _media_list_edit_worker(media_id):
         latest_scrape_job = scrape_jobs['job_states'][0]
     # active feed count
     feeds = user_mc.feedList(media_id)
-    active_syndicated_feeds = [f for f in feeds if f['active'] and f['feed_type'] == 'syndicated']
+    active_syndicated_feeds = [f for f in feeds if f['active'] and f['type'] == 'syndicated']
     active_feed_count = len(active_syndicated_feeds)
     return {
         'media_id': media_id,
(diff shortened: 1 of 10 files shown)


pypt:
After we deploy our https://github.com/berkmancenter/mediacloud/pull/421, you should be able to test it out against production because the API will be returning both the old (`feed_status`, `feed_type`) and the new (`active`, `type`) fields:

```json
{
"feed_type": "syndicated",
"type": "syndicated",
"feed_status": "active",
"active": true
}
```

...allowing for you to take your time to deploy changes on your end.

After you start using the new fields on the frontend, we'll scrap the backwards-compatible ones:

```json
{
"type": "syndicated",
"active": true
}
```

pypt:
We have deployed https://github.com/berkmancenter/mediacloud/pull/421 on our end, so please go ahead with this PR. Please let me know if it works and whether we can proceed with our final cleanup tasks.

PR #1174, opened 2018-06-20
Feature raven sentry for js

cindyloo:
maybe tweak some more depending on what we want to isolate?

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -24,12 +24,6 @@ function configDevelopmentStore(appName) {
       promiseMiddleware(),
       reduxRouterMiddleware,
       thunkMiddleware,
-      // errorReportingMiddleware,
-      createRavenMiddleware(Raven, {
-        breadcrumbDataFromAction: action => (
-          { STRING: action.str }
-        ),
-      }),
     ),
     window.devToolsExtension ? window.devToolsExtension() : f => f
   ));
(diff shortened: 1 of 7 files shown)


src/store.js
```diff
@@ -18,14 +21,26 @@ const middlewares = [
 
 function configDevelopmentStore(appName) {
   return createStore(getRootReducer(appName), {}, compose(
-    applyMiddleware(...middlewares),
+    applyMiddleware(
```
rahulbot: I don't think we want sentry logging in Dev... just when we're running in production, no?

src/store.js
```diff
@@ -18,14 +21,26 @@ const middlewares = [
...
+    ),
     window.devToolsExtension ? window.devToolsExtension() : f => f
   ));
 }
 
 function configProductionStore(appName) {
   return createStore(getRootReducer(appName), {}, compose(
-    applyMiddleware(...middlewares),
+    applyMiddleware(...middlewares,
+      createRavenMiddleware(Raven),
```
rahulbot: Why doesn't this have the `breadcrumbDataFromAction`?  Did that not work?

cindyloo: I'm not convinced it is giving us the info we want yet. What should we be sending over, and do you see what you would expect to see in the Sentry logs?

src/store.js
cindyloo: oh right, that makes sense

Issue #1148, opened 2018-06-07
click on sample story to show details about it in Explorer


rahulbot:
To show the depth of our system and metadata about stories/sources, we should add a new feature to the Explorer.  When a user click on the title of one of the sample stories, we should show a DrillDown widget with content very similar to the Topic story details page:

* `CloseIconButton` in the top right that closes the DrillDownContainer when clicked
* `AppButton` that says "Read this Story" in the top right; when clicks opens the story url in a new tab
* `<h2>Details for {storyTitle}</h2>`
* `<p>Published in <a href={source-manager-media-url}>{sourceName}</a> on {pubDate}. Written in {storyLanguage}</p>`
* one row with with tables, each table 1/4 width table (if it fits)
  * `<h3>Themes</h3>`; list of theme tags underneath
  * `<h3>Geographic Focus</h3>`; list of geographic place tags underneath
  * `<h3>People Mentioned</h3>`; list of people place tags underneath
  * `<h3>Organizations Mentioned</h3>`; list of organization tags underneath
* `<h2>Published in {sourceName}</h2>`
  * `StatBar` underneath that with the Media Source metadata (5 boxes)
* the background color of the selected story in the table should turn Explorer purple to indicate that it is the currently selected one (and go back to normal when the drill down is closed, or a different one is selected

That might be too crowded, but is a reasonable started point for getting this info out there.

rahulbot:
This requires a similar new `DrillDown` architecture to #1147, so they should be thought about together

cindyloo:
with @rahulbot updates, how do we want to do this- same way in which the QueryResultsContainer brokers the drilldown state for each drilldown?

cindyloo:
so with @rahulbot's WordInContextDrillDownContainer, the 'state' that is used to show/hide drill down is the selected data - in that case, selectedWord. Here, it would be selectedStory that would drive the showing/hiding of the drilldown

Issue #1147, opened 2018-06-07
add KWIC to Explorer when user clicks on word


rahulbot:
When the user clicks on a word like "robot" in the Top Words results widget in Explorer, we should show a new DrillDownContainer underneath with:

* CloseIconButton in the top right that closes the DrillDownContainer when clicked
* `<h2>Details "robot"</h2>`
* an `AppButton` that says "Add to all Queries and Search" in the top right
* a `WordTree` widget (which requires making a call to get sample sentence much like we do on the word detail page in Topic Mapper - `WordInContextContainer`)
* the selected word in the cloud should turn Explorer purple to indicate that it is the currently selected one (and go back to normal when the drill down is closed, or a different one is selected)

This should be styled like the existing `QueryAttentionOverTimeDrillDownContainer`, which means you might want to create a new shared `.drill-down` SCSS class and have that existing one and this new one both use it.

This will have the added benefit of making it clearer that you can add a word to all the queries and rerun.

cindyloo:
I think a drilldown HOC makes sense
will have four inputs
open function
close function
doAction what to do when something is clicked
child component - eg worddrilldown, storydrilldown, etc

cindyloo:
using .selected class to highlight the word

cindyloo:
so the remaining question is do we leave the word tree visible when searching? Now there is an explicit close button, so by default do we close it or leave it open? is it useful to keep it open or confusing?

Issue #845, opened 2018-02-12
migrate story read details to common


rahulbot:
We will soon want to show read-only information about a specific story in Sources and Explorer too. The first step in supporting this is migrating the story details in Topics right now to the common (action, reducer, and server endpoint). Story entities are already in there (`storyActions`)

This needs some cleanup because the stuff in common already for story seems to duplicate some of what is in the topic/stories folder.  I want to centralize the read-only actions, reducers, and components/containers into the common folders so all the tools can use them.

cindyloo:
StoryDetails is pretty straightforward - are we going to want to provide more information comprised by StoryContainer?

cindyloo:
to be more specific - in the cases of inlinks, outlinks and words, how will we apply to sources and explorer? if there is an id (topics_id or media_id or tags_id), generalize and pass it in via the q parameter? or am I being too broad and you just want a common Detail info...?

cindyloo:
I think my answer is in #1148- the information we want there can be fetched by common routines..

PR #1197, opened 2018-07-18
Feature refactor story details to common, includes #1147 and #1148 #845 updates

cindyloo:


cindyloo:
yes, I think I need to go back over the topic handling b/c I want inlinks, outlinks and words to be in the common story reducer although only topic stories will use it. (I think we'd use story words elsewhere though at some point perhaps)

cindyloo:
storyInfo vs topicStoryInfo- leaving the inlink and outlink reducers and handling in topics, extracted common info, including words b/c I *think* we will have a common use for the top words in stories?
also, do we need this in common storyInfo:
```
   publishDateObj: moment(payload.publish_date).toDate(),
    dateGuessMethod: tagWithTagSetsId(payload.story_tags, TAG_SET_DATE_GUESS_METHOD),
    extractorVersion: tagWithTagSetsId(payload.story_tags, TAG_SET_EXTRACTOR_VERSION),
    geocoderVersion: tagWithTagSetsId(payload.story_tags, TAG_SET_GEOCODER_VERSION),
    nytThemesVersion:
```

cindyloo:
pulling media-metadata works reasonably well..

server/views/explorer/story_samples.py
```diff
@@ -22,7 +22,10 @@
 def api_explorer_story_sample():
     solr_q, solr_fq = parse_query_with_keywords(request.args)
     story_sample_result = apicache.random_story_list(solr_q, solr_fq, 50)
-    return jsonify(story_sample_result)  
+
+    for story in story_sample_result:
+        story["media"] = apicache._mc_client().media(story["media_id"])
+    return jsonify(story_sample_result)
```
rahulbot: Two comments: 
* How much does this slow the call down? 
* Why isn't this creating its own mc_client?  Seems not good to use the apicache's *private* helper method to fetch it. Is there an `apicache.media(media_id)` call we can use to get the cached version of the media?

cindyloo: it doesn't slow it down much, once cached - if we are only getting a small set of stories

re the mc_client- that is in the explorer's apicache, so shouldn't we must make it public?

rahulbot: We should only make it public if we need to. The is already a cached media helper in apicache - `apichace.media(media_id)`.  This code, and the code you added for the demo endpoint, should call that to reduce duplicate code and make it (marginally) faster.

PR #1199, opened 2018-07-19
#815

cindyloo:


cindyloo:
note- ignore the package-lock.json commit

PR #1236, opened 2018-08-16
Feature topic (not explorer) ui updates **

cindyloo:
**notice I typed 'explorer' in the branch but meant only topic updates


cindyloo:
some formatting can be done for dates and for story text

PR #1318, opened 2018-10-04
Feature topic csv and tree

cindyloo:


cindyloo:
https://www.highcharts.com/errors/17 
The requested series type does not exist
This error happens when you are setting chart.type or series.type to a series type that isn't defined in Highcharts. A typical reason may be that your are missing the extension file where the series type is defined, for example in order to run an arearange series you need to load the highcharts-more.js file.

so, we need to pull in the highcharts-treemap

cindyloo:
I believe all concerns except for #2 have been addressed.

cindyloo:
ok I think we are ready to go
-refactored the fetching into two distinct asyncContainers

PR #1386, opened 2018-11-21
upgrade to python 3

rahulbot:
You'll need to install python 3.6.5 and make a new virtualenv:
```
pyenv install 3.6.5
pyenv virtualenv 3.6.5 mc-web-tools
```

Hooray for default unicode support :-)


cindyloo:
I think we pull this into 3.7.1 this week..

PR #1411, opened 2018-12-11
Feature story details admin

cindyloo:
one small detail - does it make sense to make a special call to download a csv for everything not included in the entity and nyt csvs or more sense to download all of it?


src/components/explorer/results/drilldowns/SelectedStoryDrillDownContainer.js
```diff
@@ -58,6 +61,7 @@ class SelectedStoryDrillDownContainer extends React.Component {
             <Row>
               <Col lg={12}>
                 <div className="actions">
+                  <AppButton variant="outlined" onClick={() => handleStoryManageClick(selectedStory)}><FormattedMessage {...localMessages.goToManageStory} /></AppButton>
```
rahulbot: This needs to have admin-only permissions around it

PR #1416, opened 2018-12-17
Explorer curly quotes fix

rahulbot:
I don't think this made it into the 3.7.0 release, because the merge from this branch into master happened before the curly quote fix was added. Should be merged and included in 3.7.1 I think.


cindyloo:
I see. Ok, we will pull into 3.7.1 along with the python update

Issue #1599, opened 2019-06-17
let admins see matching strings in cached story text


rahulbot:
It can be quite difficult to debug long queries without seeing the context of the matching phrase.  We should add some ability for admins to do this.

rahulbot:
When admins do a drill down on a story in Explorer they see a new menu of options:
![Search___Explorer___Media_Cloud](https://user-images.githubusercontent.com/673178/59605373-a364ae00-90dc-11e9-861d-194f3fca3319.jpg)
If they click the "Cache Text (highlighted)" one, then they get bumped over to the story cached text page, and any word matching the *currently selected tab in the sample stories section* get highlighted in yellow.

PR #1600, opened 2019-06-17
highlight words matching queries for admins

rahulbot:
For #1599.  This required refactoring the `selectStory` action to accept more than just and ID, so it has touchpoints in a few places. That action now accepts an object, so that we can pass the query string over to the `SelectedStoryDrillDownContainer`, which the uses the `logic-query-parser` library to translate the boolean logic query into a parsed tree we can extract strings from. The extraction isn't perfect, but I think it'll be a fine 80/20 that handles most queries.  I tried it on some reasonable complex one and it parsed pretty well.

### DIFF
--- a/src/components/common/admin/HighlightedText.js
+++ b/src/components/common/admin/HighlightedText.js
@@ -1,0 +1,26 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+// https://stackoverflow.com/questions/29652862/highlight-text-using-reactjs
+
+const HighlightedText = ({ text, search }) => {
+  if (search === undefined) {
+    return text;
+  }
+  const regex = new RegExp(`(${search.join('|')})`, 'gi');
+  const parts = text.split(regex);
+  return (
+    <span>
+      {parts.filter(part => part).map((part, i) => (
+        regex.test(part) ? <mark key={i}>{part}</mark> : <span key={i}>{part}</span>
+      ))}
+    </span>
+  );
+};
+
+HighlightedText.propTypes = {
+  text: PropTypes.string.isRequired,
+  search: PropTypes.array,
+};
+
+export default HighlightedText;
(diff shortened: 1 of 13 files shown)


src/components/explorer/results/drilldowns/SelectedStoryDrillDownContainer.js
```diff
@@ -25,7 +27,34 @@ const localMessages = {
...
   published: { id: 'explorer.story.published', defaultMessage: 'Published in {media}' },
-  goToManageStory: { id: 'drilldown.inContext.title', defaultMessage: 'Manage Story...' },
+  adminOptions: { id: 'drilldown.adminOptions.title', defaultMessage: 'Admin Options...' },
+  goToManageStory: { id: 'drilldown.adminOptions.manageStory', defaultMessage: 'Manage Story' },
+  highlightedCachedText: { id: 'drilldown.adminOptions.highlightedCachedText', defaultMessage: 'Cached Text (highlighted)' },
+};
+
+const extractStringsFromParseTree = (node) => {
+  let words = [];
+  if (node.lexeme.type === 'string') {
```
cindyloo: didn't know we could inspect the node in this way. nice!

src/components/common/admin/HighlightedText.js
```diff
@@ -0,0 +1,26 @@
...
+const HighlightedText = ({ text, search }) => {
+  if (search === undefined) {
+    return text;
+  }
+  const regex = new RegExp(`(${search.join('|')})`, 'gi');
+  const parts = text.split(regex);
+  return (
+    <span>
+      {parts.filter(part => part).map((part, i) => (
+        regex.test(part) ? <mark key={i}>{part}</mark> : <span key={i}>{part}</span>
```
cindyloo: also didn't know about this html element!

hroberts:
this is the code that matches the topic pattern against story content for relevancy checking:

re2.search(topic['pattern'], content, re2.I | re2.X | re2.S)

we use re2 because checking complicated patterns over large chunks of text can rarely cause the standard re to take 20+ minutes to complete.  if you are just running against individual sentences, this shouldn't be a problem.

I'm worried that this won't work well, though, for ANDed queries.  if you are searching for 'trump and russia' and just searching for sentences that match 'trump.*russia|russia.*trump', you will often find no sentences that match since the two words are likely not to appear in the same sentence.  I had to implement the inclusive pattern matching to handle this.  I could make that boolean query -> inclusive regex translation code available as a stand alone module to you, or (cleaner I think) just generate the inclusive pattern and stick it into topic.inclusive_pattern.

PR #1706, opened 2020-01-01
Add syntax highlighting to explorer keywords input

rahulbot:
Strings show up in purple, operators in black, field names in blue, numbers in green (see attached image). This one should be checked with the research team to see how they feel about it.

Search___Explorer___Media_Cloud


hroberts:
looks great!  there may be some very small differences between how the query code is being parsed and executed on the back end and how this package is doing the syntax highlighting, but that's always the case with syntax highlighting.  anything to give the user more feedback to get the query correct should be very helpful.

PR #1792, opened 2020-04-13
invalid CSS

rleir:
Hi all, 
This CSS has a benign problem because these settings are overridden by other CSS. So, the change does not fix anything affecting the user. However, I like to tidy up my Deepscan so that actual bugs will not get missed.
"Condition '`${d.belowTextColor} !important`' is always true because it evaluates to a non-empty string."
These changes have not been tested, because I have not figured out the build system (embarrassed), so feel free to ignore this PR.
thanks -- Rick


rleir:
> Hi - thanks for the PR! I'm curious how you came onto this code - are you using these widgets as part of a different project, or experimenting with the larger MediaCloud front-end codebase?

Hi Rahul
I have a few free cycles, and want to help. But first I need to learn how everything here works, and one thing I did was run the Deepscan linter. 


> These should be two easy fixes for you to make.

And ESlint is teaching me a better coding style, thanks!.
cheers -- Rick

rleir:
This syntax is less confusing, and I think the linter will be ok with it.

rleir:
I want to clean this PR up. You will see a new one shortly.

PR #1799, opened 2020-04-17
Dev start doco

rleir:
Here is a first cut at the Ubuntu front-end dev doco, just so you can see which direction I am going. Comments, please!


rleir:
No, this PR includes extraneous changes from other work. Now, how to clean it up ...

PR #1801, opened 2020-04-18
Dev ubuntu doc

rleir:
Rahul,
For some reason, the changes for both PR's are appearing here. I put the changes in separate branches, so they should be separated in the PR's. It is a mystery, let me think about it.
thanks
Rick


rleir:
Ok, I have a branch on a branch. I wanted two branches from master. D'oh, next time I will change back to master before creating a branch.

PR #1802, opened 2020-04-18
a first cut at dev doco on Ubuntu

rleir:


rleir:
Hi Rahul
Now the nodejs app is failing with CSP console errors: "Refused to load the font '<URL>' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'font-src' was not explicitly set, so 'default-src' is used as a fallback." What am I doing wrong?
thanks -- Rick

rleir:
Fixed, by 'mkdir build/public'.  The 'public' contents is not getting built, that is my next challenge. There should be an img dir tree containing sample searches and a gen dir containing the app.

PR #1826, opened 2020-05-13
Update config readme

dsjen:
@rahulbot / @cindyloo -- `config/server.config.template` does not exist. Is the proposed change in this PR correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Coming soon
 Configuration
 ------------- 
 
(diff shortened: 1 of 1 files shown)


cindyloo:
I believe this means the config/app.config.template

dsjen:
Ok, looks  like it should be `app.config` rather, referenced in https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/bfa5a2b8941db84df375f5646c8a79b0c9742377/server/util/config.py#L61

PR #1834, opened 2020-05-19
Add start redis to readme

dsjen:


dsjen:
oh wow, I'm running 5 actually! It seems to work, but I'm not confident that I've hit it as I've only logged in to see if things work.

PR #1835, opened 2020-05-19
Update instructions for installing MongoDB

dsjen:
I updated this to just link to the mongo instructions (the original `brew` ones are outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Node and npm:
  * `npm install` to install all the package dependencies (as specified in the `package.json`)
 
 MongoDB:
-[Install MongoDb](https://docs.mongodb.com/manual/installation/).  We develop on OS X and install via the [HomeBrew package manager](http://brew.sh): `brew install mongodb`
+[Install MongoDb](https://docs.mongodb.com/manual/administration/install-community/).
 
 Redis:
 [Install Redis](http://redis.io/)  We develop on OS X and install via the [HomeBrew package manager](http://brew.sh): `brew install redis`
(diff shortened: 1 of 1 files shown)


dsjen:
Yes, I went with mongo 4.2. I haven't don't done much more than login at the moment, so am not super confident that I've hit mongo in anything I've done.

dsjen:
Gotcha! Created #1839 for upgrading mongo

dsjen:
`brew` works, but the brew command has changed. `brew install mongodb` no longer works and it's replaced with something like `brew install mongodb-community@4.2`. We can certainly stick that in the readme for convenience. I just figure that it's changed once and it's going to change again with ever new version, so why not just link to the instructions instead in this case.

dsjen:
Hmm, everything went swimmingly for me, but I'm on OS 10.14.5. Maybe we stick to the older version of mongo for now?

dsjen:
Just to clarify, I believe our two questions are:
* Should we upgrade mongodb to 4.2?
* Should we include the exact instructions installing mongodb via brew?

I'd vote to upgrade. I don't feel strongly about whether we include the exact instructions, but lean toward linking to the page that describes how to install.

cindyloo:
I have mongo 3.2.10 and will evaluate the update now..

cindyloo:
I found it straightforward to install:
`brew tap mongodb/brew`
then
`brew install mongodb-community@4.2`
as documented
[here](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/)

cindyloo:
@rahulbot did we make a dump/migration mechanism to preserve user info within the mongodb?

dsjen:
ok, updated to include the brew instructions!

PR #1836, opened 2020-05-19
Removed JS tests

### LINKED ISSUE #1829: remove really old JS tests
Back in #172 we started doing some JS testing. We have retired this in favor of Python based test (#1690). As part of our "summer cleaning" I think we should:
1. remove the `test` dir that has that really old JS code
2. remove the related dependencies in `package.json` (`nightmare` springs to mind, but are there others?)

dsjen:
Fixes #1829


test/helpers/visit.js
```diff
@@ -1,49 +0,0 @@
-import nightmare from 'nightmare';
-import url from 'url';
```
dsjen, line 2: FYI, I never found the `url` package, so I assume that it has been removed already.

PR #1841, opened 2020-05-20
Added SVG download for Explorer attention bubble chart

### LINKED ISSUE #1830: add SVG download to explorer total attention widget
You can't download an SVG of the bubble chart of total attention. That seems like an oversight. Let's add support for downloading that chart.

Search___Explorer___Media_Cloud

Rough process I would suggest:
1. Review `QueryTotalAttentionResultsContainer.render` - note that for the `downloadOptimizer` it renders a `AppButton` if there is only one query but renders an `ActionMenu` if there are multiple queries
2. Refactor it to render a `ActionMenu` no matter how many queries there are
3. add a new `MenuItem` to the download `ActionMenu` to allow the user to "Download a SVG"

We have a generic helper for downloading SVG, called `downloadSvg` in `util/svg`. You pass it the filename you'd like to use and the dom id (or node) of the `<SVG>` tag to download. So you'll need to:
```js
import { downloadSvg } from '../util/svg';
```
and then in the `onClick` handler of the `MenuItem` call that with the id of the `BubbleRowChart`, which is already in a constant as `BUBBLE_CHART_DOM_ID`. In theory that should be it!

dsjen:
Fixes #1830

Screen Shot 2020-05-20 at 4 27 10 PM

### DIFF
--- a/src/components/explorer/results/QueryTotalAttentionResultsContainer.js
+++ b/src/components/explorer/results/QueryTotalAttentionResultsContainer.js
@@ -12,6 +12,7 @@ import AppButton from '../../common/AppButton';
 import ActionMenu from '../../common/ActionMenu';
 import BubbleRowChart from '../../vis/BubbleRowChart';
 import { queryChangedEnoughToUpdate, postToDownloadUrl, ensureSafeResults } from '../../../lib/explorerUtil';
+import { downloadSvg } from '../../util/svg';
 import messages from '../../../resources/messages';
 import { FETCH_INVALID } from '../../../lib/fetchConstants';
 import withQueryResults from './QueryResultsSelector';
@@ -28,6 +29,7 @@ const localMessages = {
   },
   downloadOneCsv: { id: 'explorer.attention.total.downloadCsv', defaultMessage: 'Download all story URLs' },
   downloadCsv: { id: 'explorer.attention.total.downloadCsv', defaultMessage: 'Download all story URLs for {name}' },
+  downloadSvg: { id: 'explorer.attention.total.downloadSvg', defaultMessage: 'Download SVG' },
   viewNormalized: { id: 'explorer.attention.mode.viewNormalized', defaultMessage: 'View by Story Count (default)' },
   viewRegular: { id: 'explorer.attention.mode.viewRegular', defaultMessage: 'View by Story Percentage' },
 };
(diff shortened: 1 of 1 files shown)


src/components/explorer/results/QueryTotalAttentionResultsContainer.js
```diff
@@ -28,6 +29,7 @@ const localMessages = {
   },
   downloadOneCsv: { id: 'explorer.attention.total.downloadCsv', defaultMessage: 'Download all story URLs' },
   downloadCsv: { id: 'explorer.attention.total.downloadCsv', defaultMessage: 'Download all story URLs for {name}' },
+  downloadSvg: { id: 'explorer.attention.total.downloadSvg', defaultMessage: 'Download SVG' },
```
dsjen, line 32: Do I need to do anything to to extract this string for i18n?

dsjen:
Also, I made this PR against master. Should I make it against `v3.14.x` instead?

src/components/explorer/results/QueryTotalAttentionResultsContainer.js
rahulbot: That syntax correct. However many repetitive strings like this are stored in `src/resources/messages.js`. For instance, you'll see there is a 'Download a SVG' string in there that would work for this usage. In cases like this you simple make sure to `import messages from '../../../resources/messages';` and then you can simply use `messages.downloadSVG` as your string where you need to in `render`. No new `localMessages` item required 👍🏽

dsjen: I see! Thanks for the suggestion! Just so that I know, if I were to add a new string, do I need to extract them for translations or does this just automatically happen somehow?

rahulbot: We don't need to extract them for usage anywhere - it just uses the defaultMessage.  Ie. we have the architecture in place to localize, but haven't localized (not even to US-EN).

dsjen:
Thanks! I think that pattern makes sense (patches to the current release branch) and new features merging to master. This is probably something to discuss outside of a PR, but that you mention "long time" between a release makes me wonder if we can figure out or invest in strategies to shorten that cycle.

src/components/explorer/results/QueryTotalAttentionResultsContainer.js
dsjen: Out of curiosity, is there a plan to localize? Why go through the effort in the first place (not a judgement--just trying to understand)?

PR #1850, opened 2020-05-26
Replace dangerouslySetInnerHTML with SanitizedHTML

### LINKED ISSUE #1768: replace any usages of dangerouslySetInnerHTML with SanitizedHTML
I added the `<SanitizedHTML>` helper as a dependency of our project. We need to go back and refactor our code so that any existing usages of `dangerouslySetInnerHTML` are switched to use this new helper instead.

Docs on how to use it are here: https://www.npmjs.com/package/react-sanitized-html

Related commit: 74d56b0e96922a0c1b13f48630de2dae19d21e06

dsjen:
Fixes #1768

### DIFF
--- a/src/components/common/news/RecentNewsItem.js
+++ b/src/components/common/news/RecentNewsItem.js
@@ -13,7 +13,7 @@ const RecentNewsItem = props => (
       </Col>
       <Col lg={8}>
         <SanitizedHTML html={props.item.note} />
-        <SanitizedHTML html={props.item.details} />
+        {props.item.details && <SanitizedHTML className="news-details" html={props.item.details} />}
       </Col>
       <Col lg={1}>
         <span className="item-app-list">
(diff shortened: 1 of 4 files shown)


src/components/common/news/RecentNewsItem.js
```diff
@@ -11,8 +12,8 @@ const RecentNewsItem = props => (
         <span className={`news-type news-${props.item.type.toLowerCase()}`}>{props.item.type}</span>
       </Col>
       <Col lg={8}>
-        <span dangerouslySetInnerHTML={{ __html: props.item.note }} />
-        <small dangerouslySetInnerHTML={{ __html: props.item.details }} />
+        <SanitizedHTML html={props.item.note} />
+        <SanitizedHTML html={props.item.details} />
```
rahulbot, line 16: I think this lost a `<small>` tag here, no?

dsjen: Good catch!

dsjen:
Updated, e.g.:

Screen Shot 2020-05-26 at 8 15 52 PM

Turns out that I'm unable to use the `<small>` tag as `SanitizedHTML` wraps everything in a dev, so the small styles didn't apply. I changed it to CSS instead.

PR #1860, opened 2020-06-02
Added Makefile

dsjen:
Would either of you be opposed to a `Makefile`? I will just never remember the right incantations for installing requirements and I think this would be a nice shortcut we could use for consistent installs. We can also add other commands that we want to standardize.

I'll be looking for both of you to agree to this before merging and will add to the docs (just didn't want to invest in that before you agree to adding a `Makefile`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+requirements.py:
+	pip install -q -r requirements.txt --exists-action w
+


dsjen:
Yeah... I think the benefit of this is that we can codify some of the standard stuff (e.g. when we get around to python linting we can “make lint”) and possibly make automation slightly more straightforward.

dsjen:
@cindyloo -- do you have any opinions on this?

PR #1861, opened 2020-06-02
Removed password reset message

dsjen:


dsjen:
Is the next step to cherry-pick this change to the versioned branch/tag?

dsjen:
Ah yes, I remember that process now. I think I've perhaps done this opposite in that I made the change on master and am wanting to bring this change over to the versioned branch for deployment. Would you be ok with that?

Just to be concrete, I'm proposing:
* Cherry pick this commit into `v3.13.x`
* Tag it as `topics-v3.13.20`
* Deploy to all dev instances
* Confirm it's working
* Deploy to all prod instances

dsjen:
Gotcha! I'll edit the plan I outlined then. Do I want for the deploy to go to prod instead of dev though? Is dev being used for the topics stuff?

dsjen:
Ok, sounds good!

PR #1864, opened 2020-06-04
Switch to new query data helper #1862

### LINKED ISSUE #1862: switch all api client date helper calls
As per notes in https://github.com/mitmedialab/MediaCloud-API-Client/issues/73, we should switch all the calls to `publish_date_query` in this codebase to be start/end date inclusive by upgrading to client v3.12.0 and using the new `dates_as_query_clause` helper.

rahulbot:
This small change fixed #1862. This is mostly for cleanup, but does effect some story counts shown to users in non-research-critical places:
* source manager attention charts
* topic version seed query totals

Note that bumps the mediacloud-api library dependency up, so you'll have to `pip install` to get the latest library for it to work. Let me know how it looks and I'll merge.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 Flask==1.1.2
 Flask-Login==0.5.0
 Flask-Webpack==0.1.0
-mediacloud==3.11.3
+mediacloud==3.12.0
 pymongo==3.10.1
 requests==2.23.0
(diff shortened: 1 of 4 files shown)


cindyloo:
I'll run it in the branch tomorrow unless you want to integrate and then we can test the overall re-factoring

PR #1867, opened 2020-06-04
Document bug fix & patch release process

dsjen:
I wanted to record the bug fix / patch process as it wasn't documented anywhere.

### DIFF
--- a/doc/releases.md
+++ b/doc/releases.md
@@ -6,8 +6,8 @@ We release our web-apps in containers on a server running the [Dokku PaaS](http:
 One-Time Setup
 --------------
 
-To be able to release, you first need to setup your keys so that you can push to the Dokku server. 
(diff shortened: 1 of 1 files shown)


doc/releases.md
```diff
@@ -27,13 +27,21 @@ Building and Pushing a Release
...
 10. Push the tag(s) "git push origin --tags"
 <br/><br/>
 11. Create a PR on GitHub from your banch back to master (so the changes get merged in)
+
+Bug Fixes and Patches
+---------------------
+
+When fixing a bug for production
+* Make a commit to fix the bug on the release branch (e.g. v3.13.x)
+* Cherry-pick the commit to master
```
rahulbot, line 46: Step 11 on the list above notes that we should make a PR back to master after the release, so why is the idea of cherry-picking back to master needed here?

dsjen: Good point! I'll remove this point

PR #1869, opened 2020-06-05
Refactored CSV validation and saving

### LINKED ISSUE #1773: refactor any URL downloading to use new `server.util.file` helper
We probably have a handful of places in our server code where we download content from a remote URL to save locally temporarily. The one that comes to mind is the CSV download feature we support in `topic.platforms.generic_csv.py`.

We should refactor those to use the new `server.util.file.save_url_to_temp_file` helper so this process is implemented in just one place.

dsjen:
Fixes #1773

I tested this in the topics multiplatform csv upload and in the sources upload.

### DIFF
--- a/server/util/file.py
+++ b/server/util/file.py
@@ -6,6 +6,12 @@ from werkzeug.utils import secure_filename
 from server import app
 
 
+def save_file_to_upload_folder(file_to_save, filename):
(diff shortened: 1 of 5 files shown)


dsjen:
I'm a decorator nut!

Issue #1413, opened 2018-12-12
set up python linting


rahulbot:
Lets copy https://github.com/mitmedialab/gobo/pull/81

cindyloo:
I'll be learning from you! I was new to linting before MC so I'm eager to learn some other integrations as well!

rahulbot:
Sounds smart to phase it in. Also consider how this can integrate with PyCharm so the warnings/errors match.

dsjen:
Yes, there’s always so much to learn from each other!

I’ll document or at least add pointers for pycharm integration too!

PR #1870, opened 2020-06-09
Added python linting

dsjen:
Part of #1413

* Separated prod vs. local requirements so that pylint and testing packages aren't dependencies on prod. Install `requirements.txt` will install the prod requirements still.
* Moved requirements into it's own directory
* Added `lint.py` command to `Makefile`
* Added instructions to README
* Added autogenerated `.pylintrc`

`.pylintrc` contains the linting rules. It's mostly autogenerated, except that I've added any rule that was violated when running the linter on our `server` files to the `disable=` line, on line 63. You'll find a "TODO" comment for where I started adding these rules. The one exception is `missing-docstring` which I added directly above the "TODO" because I think we don't need to care about that rule.

Moving forward, we can chip away at enabling these rules or moving them above the line because we don't care to abide by them. Ones that I find interesting are:
* useless-import-alias,
* superfluous-parens,
* deprecated-method,
* dangerous-default-value,
* unsubscriptable-object,
* reimported,
* redefined-outer-name,
* using-constant-test,
* useless-object-inheritance,
* super-init-not-called,
* pointless-statement,
* inconsistent-return-statements,
* cyclic-import

After merging this:
* I will be to set up Travis to run the linter whenever a pull request is made
* We can then tackle specific rules by first deciding if we care. If we do, then we can fix the errors when enabling the rule.


dsjen:
Neat trick w/ the requirements, huh?!?

I'm only on board with making sure that we add linting for what matters only. We don't want it to get in the way.

PR #1872, opened 2020-06-09
Added Travis config

### LINKED ISSUE #1413: set up python linting

dsjen:
Fixes #1413 (we'll still need to enable the rules, but lets break that out into separate granular tasks).

Check out the checks that happen with this PR:
Screen Shot 2020-06-09 at 12 24 06 PM

We can also think about running tests as part of the process too. I  ran them locally and a bunch of them failed though.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -202,7 +202,8 @@ disable=print-statement,
         invalid-name,
         unused-variable,
         duplicate-code,
-        cyclic-import
+        cyclic-import,
+        undefined-variable
 
 # Enable the message, report, category or checker with the given id(s). You can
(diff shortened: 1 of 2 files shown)


.pylintrc
```diff
@@ -202,7 +202,8 @@ disable=print-statement,
         invalid-name,
         unused-variable,
         duplicate-code,
-        cyclic-import
+        cyclic-import,
+        undefined-variable
```
dsjen, line 206: I guess I somehow missed disabling this rule (it was failing on travis)

Issue #1638, opened 2019-08-09
link custom collection to advanced source search page


rahulbot:
People will want to see the list of sources matching their custom collections - we should link the custom collection widget to the advanced source search results page on Source Mgr. This will let people page through the matching sources (and ideally let them download that list as well).   

![Search___Explorer___Media_Cloud](https://user-images.githubusercontent.com/673178/62792544-24dd2900-baed-11e9-8142-5ae2da52b8d9.jpg)

I'm pretty sure this is blocked by #1637, because we can't encode a custom collection over to that page in a URL (and it doesn't support multiple selections within a metadata filter).

cindyloo:
here is another Custom Collection implementation issue....

rahulbot:
Yep. My thought here seems to have been to (on #1637) we use some kind of wrapper that parses the URL-encoded "custom collection criteria" and then passes that in as `initialValues` to the media-picker-based advanced search page.

cindyloo:
within MediaPicker,  upon the clicking of `Add as Custom Collection` we need to do two things:
- store all the media_ids that are part of this selection
- show them all as selected
Then, when the "OK" button is clicked, we serialize those media_ids into the URL as we do for sources and collections, along with the `media_keyword` and the metadata tags. That way, we can utilize the custom collection mechanism outside of the MediaPicker

rahulbot:
Can you explain that approach more? I like how you architected it when you first implement it; having a "custom collection" just be a list of the media metadata tags the user selected (ie. the search criteria). Why do we need to turn that into a list of media to support using "custom collections" in other places?

I think this particular issue, which probably isn't high priority, would entail:
1. Adding a click handler to the custom collection rendering
2. When clicked, finding a way to serialize the selected media metadata search into a URL and appending that to the link over to advanced search page
3. Parsing that serialized data on the advanced search page
4. Passing the parsed "custom collection" into the media picker as preselected search criteria for the drop down menus

cindyloo:
right, we don't want or need to serialize all media_ids into the URL - we want to special case the search + media_keyword + metadata tag in python to insert those media ids into the solr search query when needed eg `media_id: {[...]}`.
we do want to display correlated media as selected in MediaPicker display page

cindyloo:
fyi the Custom Collection handling is currently broken, which makes this a more timely consideration

rahulbot:
Ah. I hear two considerations here:

**1)** How do we search in Explorer with a custom collection?
[...]

PR #1875, opened 2020-06-09
Feature sources media picker integration

cindyloo:
flat layout for sources - as accessed by clicking "Advanced Search" in search field
preserve dialog for topics/explorer

-maybe change default # of rows coming back?


src/styles/sources/_search.scss
```diff
@@ -1,11 +1,88 @@
...
-        position: absolute;
-        right: 25px;
-        top: 10px;
+@import '../_constants';
+
+.select-media-sources {
+
+    .select-media-sidebar-sources {
+      float: left;
+      width: 230px;
```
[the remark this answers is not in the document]
cindyloo: when you say moved into a variable, what do you mean? into another css style or ?

src/components/source/search/MediaPickerSourceSearchContainer.js
```diff
@@ -0,0 +1,113 @@
...
+import { Grid } from 'react-flexbox-grid/lib';
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import withAsyncData from '../../common/hocs/AsyncDataContainer';
+import PickedMediaContainer from '../../common/mediaPicker/PickedMediaContainer';
+import MediaPickerResultsContainer from '../../common/mediaPicker/MediaPickerResultsContainer';
+import { initializePreviouslySelectedMedia, clearSelectedMedia, resetMetadataShortlist, selectMediaPickerQueryArgs } from '../../../actions/systemActions';
+import { ALL_MEDIA } from '../../../lib/mediaUtil';
+
+class MediaPickerSourceSearchContainer extends React.Component {
```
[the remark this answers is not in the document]
cindyloo: the `UNSAFE_componentWillReceiveProps` and the `shouldComponentUpdate` are indeed the same and could be contained w/in a wrapper for each component. We do however need two different components for the Dialog invocation versus the Sources invocation

server/views/media_search.py
```diff
@@ -1,17 +1,39 @@
...
+from server.auth import user_mediacloud_client, user_admin_mediacloud_client
+from flask import jsonify, request
 
 logger = logging.getLogger(__name__)
 
-MAX_SOURCES = 40
+MAX_SOURCES = 60
 
+def media_search_with_page(search_str, tags_id=None, **kwargs):
+    link_id = request.args.get('linkId') if 'linkId' in request.args else 0
```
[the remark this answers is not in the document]
cindyloo: I think in this case it is best to have two different calls as the media vs collection calls are by separated throughout our code base.

src/styles/sources/_search.scss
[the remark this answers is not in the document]
dsjen: In `scss`, variables can be declared, e.g. `$sidebar-width: 230px`, which can then referenced here (e.g. `width: $sidebar-width`) and elsewhere. You can even do arithmetic with it!

src/components/source/search/MediaPickerSourceSearchContainer.js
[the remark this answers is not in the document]
dsjen: I see! The exact code is used three times, so pulling it out into a helper, extending it, etc. would certainly help w/ reusability.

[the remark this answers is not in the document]
cindyloo: creating a MediaPickerSidebarComponent now. It seems it should be a helper vs hoc - I'll push it shortly and you can help me decide if that should be the case

cindyloo:
@rahulbot  - I set up the MediaPickerSourceContainer with an async search thinking that if the user arrives at the Advanced Search with a `searchstring`, we would want to automate the search with the search parameter coming in to the Source search.  if the search string is empty, if we think those default results are not helpful, we could not do the `fetchAsync` search...

cindyloo:
> I pulled in the latest commits and found some more items that didn't work as I expected:
> 
> 1. **EditOpenWebForm**: `EditOpenWebForm.js` was not compiling. I fixed the JS error so it compiles and pushed, but have not tested in Topic Mapper to see if that widget (to the Open Web / MC platform addition screen) now works.
> 2. **None Selected**: In Explorer, when I open media picker and I remove all the selected sources/collections I expect it to work the same as previously (ie. show a "no media" item and leave the title and media showing). Instead, it seems to remove the whole widget and move the buttons up. This makes is visually confusing to me, and makes it harder to pick the "all media" option because the 3-dot menu disappears.

super  helpful - I had if/def'd here to avoid errors with no selectedMedia. Fixed.
>    Search___Explorer___Media_Cloud_and_Search___Explorer___Media_Cloud
> 3. **First Advanced Search**: When I first visit the Source Mgr advanced search page, via selecting it from the Source Mgr menu, I just see a spinning ball. If I look at my local server log I can see that it seems to be running a search right away even though I haven't done anything except come into the page (`http://localhost:5000/#/search`). This search takes a really long time (30+ seconds), so it makes me think the page isn't working at all. We need to prevent this from searching unless you click the search or "get more results" buttons.
> 
see below comment
> ```
> [16:35:08][DEBUG] mediacloud.api api.py:_query:416 | query GET to https://api.mediacloud.org/api/v2/media/list with {'last_media_id': '0', 'rows': 100, 'include_dups': 0, 'sort': 'num_stories', 'fq': 'media_source_tags: {tag_sets_id: [5, 1959, 15765102]}', 'link_id': 0} and None
> ```
> 
> 1. **Collection Search**: when I switch over to the "search collections" tab and search for "India" it works great (68 results). However when I then search for "Ghana" I still see the list of Indian sources. If I look in the browser "network" tab I can see that "Ghana" was sent to the server, and Ghanian collections came back from the server. It looks like the reducer is handling the `FETCH_MEDIAPICKER_COLLECTION_SEARCH_RESOLVED` incorrectly.
very helpful - needed to pull id from *either* media_id or tags_id
>    Source_Manager___Media_Cloud
> 2. **Sidebar**: So much better with this default tab behavior here! However I still think the `position:fixed` on `.select-media-sidebar-sources` should be removed when the media picker is in "embedded" mode like this. I recognize that is kind of a nitpick.

cindyloo:
all but one identified bug fixed - the geo collections still not clearing out

server/views/media_picker.py
```diff
@@ -34,16 +34,19 @@ def api_mediapicker_source_search():
     search_str = request.args['media_keyword']
     cleaned_search_str = None if search_str == '*' else search_str
     querying_all_media = False
+    link_id = request.args['link_id'] if 'link_id' in request.args else 0
```
dsjen, line 37: I  just realized that this could be simplified as: `request.args.get('link_id', 0)`

cindyloo: that's default behavior of the get method? if it's there, use that value, otherwise default to 0? we use the if else everywhere

dsjen: Yes, that should be the case (see: https://docs.python.org/3/library/stdtypes.html#dict.get). I think we could use this syntax in a bunch of places!

rahulbot: Super cool! I didn't know that existed. So much more succinct!
However I think `link_id` should be set to `None` if not defined (so it doesn't get sent to the back-end at all by the MC API helper).

dsjen:
I have this working locally! 🎊 I've tried not to repeat issues already mentioned, and here's my list:

* I found the "Get more results" button confusing. I kept clicking it and unsure of what was supposed to happen. I now see that it loads more results (duh!) and that it's supposed to append. Also, for my search term, there were none to append and the button just went away instead, which is also a confusing
* Looks like there's a `<br />` after the result count and keyword, which doesn't seem like it should go there:
Screen Shot 2020-06-17 at 4 09 01 PM
* I see an error when using the "Get more results" button:

```
Uncaught TypeError: Cannot read property 'setState' of undefined
    at setQueryFormChildDialogOpen (EditOpenWebForm.js:32)
    at MediaPickerDialog.eval [as handleModifyClick] (MediaPickerDialog.js:116)
    at onClick (MediaPickerDialog.js:255)
    at HTMLUnknownElement.callCallback (react-dom.development.js:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
    at invokeGuardedCallback (react-dom.development.js:440)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:454)
    at executeDispatch (react-dom.development.js:584)
    at executeDispatchesInOrder (react-dom.development.js:609)
    at executeDispatchesAndRelease (react-dom.development.js:713)
```

* Strange floating "0" appears when clearing out all media

Screen Shot 2020-06-17 at 4 19 01 PM

cindyloo:
> I pulled in the latest commits and found some more items that didn't work as I expected:
> 

> 1. **EditOpenWebForm**: `EditOpenWebForm.js` was not compiling. I fixed the JS error so it compiles and pushed, but have not tested in Topic Mapper to see if that widget (to the Open Web / MC platform addition screen) now works.
Confirmed fixed in Topic Mapper

> 2. **None Selected**: In Explorer, when I open media picker and I remove all the selected sources/collections I expect it to work the same as previously (ie. show a "no media" item and leave the title and media showing). Instead, it seems to remove the whole widget and move the buttons up. This makes is visually confusing to me, and makes it harder to pick the "all media" option because the 3-dot menu disappears.
>    Search___Explorer___Media_Cloud_and_Search___Explorer___Media_Cloud
Confirmed fixed - some  if/else checking I had inserted was removing the widget altogether


> 3. **First Advanced Search**: When I first visit the Source Mgr advanced search page, via selecting it from the Source Mgr menu, I just see a spinning ball. If I look at my local server log I can see that it seems to be running a search right away even though I haven't done anything except come into the page (`http://localhost:5000/#/search`). This search takes a really long time (30+ seconds), so it makes me think the page isn't working at all. We need to prevent this from searching unless you click the search or "get more results" buttons.
@rahulbot see below my comments/ thoughts
> 
> ```
> [16:35:08][DEBUG] mediacloud.api api.py:_query:416 | query GET to https://api.mediacloud.org/api/v2/media/list with {'last_media_id': '0', 'rows': 100, 'include_dups': 0, 'sort': 'num_stories', 'fq': 'media_source_tags: {tag_sets_id: [5, 1959, 15765102]}', 'link_id': 0} and None
> ```
> 
> 1. **Collection Search**: when I switch over to the "search collections" tab and search for "India" it works great (68 results). However when I then search for "Ghana" I still see the list of Indian sources. If I look in the browser "network" tab I can see that "Ghana" was sent to the server, and Ghanian collections came back from the server. It looks like the reducer is handling the `FETCH_MEDIAPICKER_COLLECTION_SEARCH_RESOLVED` incorrectly.
Fixed  - CountryCollections has been refactored away so I made the code work for both the Sources Collection search and the Geo Collections search

>    Source_Manager___Media_Cloud
> 2. **Sidebar**: So much better with this default tab behavior here! However I still think the `position:fixed` on `.select-media-sidebar-sources` should be removed when the media picker is in "embedded" mode like this. I recognize that is kind of a nitpick.

cindyloo:
> I have this working locally! 🎊 I've tried not to repeat issues already mentioned, and here's my list:
> 
> * I found the "Get more results" button confusing. I kept clicking it and unsure of what was supposed to happen. I now see that it loads more results (duh!) and that it's supposed to append. Also, for my search term, there were none to append and the button just went away instead, which is also a confusing

This button is now disabled if there are no more results to fetch OR if a search hasn't happened yet. we could potentially hide/show it in the latter case but I know a UI principle is not to have elements disappear/reappear

> * Looks like there's a `<br />` after the result count and keyword, which doesn't seem like it should go there:
> 
> Screen Shot 2020-06-17 at 4 09 01 PM
> 
> * I see an error when using the "Get more results" button:
> ```
> Uncaught TypeError: Cannot read property 'setState' of undefined
>     at setQueryFormChildDialogOpen (EditOpenWebForm.js:32)
>     at MediaPickerDialog.eval [as handleModifyClick] (MediaPickerDialog.js:116)
>     at onClick (MediaPickerDialog.js:255)
>     at HTMLUnknownElement.callCallback (react-dom.development.js:336)
>     at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
>     at invokeGuardedCallback (react-dom.development.js:440)
>     at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:454)
>     at executeDispatch (react-dom.development.js:584)
>     at executeDispatchesInOrder (react-dom.development.js:609)
>     at executeDispatchesAndRelease (react-dom.development.js:713)
> ```
> 
> * Strange floating "0" appears when clearing out all media
> 
> Screen Shot 2020-06-17 at 4 19 01 PM

cindyloo:
@dsjen re: the reason for <br /> tag - if the query is quite long, it wraps and looks weird. open to suggestions!

dsjen:
Thank you! I pulled the latest changes (I think you mentioned that this is ready for re-review?).

Issues that I'm seeing:
* Still seeing `Cannot read property 'setState' of undefined`
* Still seeing strange floating "0" appears when clearing out all media
* I noticed when clearing out all sources that the `x` button persists for the last source that I cleared out:
Screen Shot 2020-06-23 at 11 41 38 AM

I have a suggestion that I think helps to address my complaint about the `<br/>` as well as the `Get more results` button:
Screen Shot 2020-06-23 at 11 27 08 AM
* Can we move the get more results button to be above the table? I think it seems more natural there and less confusing than next to the search button.
* Can we change the font style of the search term and add in a colon before it? I think that will address the need of the line break.

cindyloo:
> Thank you! I pulled the latest changes (I think you mentioned that this is ready for re-review?).
> 
> Issues that I'm seeing:
> 
I had missed fixing these two -- will do
> * Still seeing `Cannot read property 'setState' of undefined`
> * Still seeing strange floating "0" appears when clearing out all media

**strange! I haven't seen this before will investigate**

> * I noticed when clearing out all sources that the `x` button persists for the last source that I cleared out:
>   Screen Shot 2020-06-23 at 11 41 38 AM
> 
> I have a suggestion that I think helps to address my complaint about the `<br/>` as well as the `Get more results` button:
> Screen Shot 2020-06-23 at 11 27 08 AM
> 

 **great suggestions! The only think about colons is that we haven't really been using them. I am not opposed here as it does look better all on the same line/wrap. The font changes will look better too**

> * Can we move the get more results button to be above the table? I think it seems more natural there and less confusing than next to the search button.
> * Can we change the font style of the search term and add in a colon before it? I think that will address the need of the line break.

dsjen:
Great, thanks! Also, if those UI changes are tricky to make, please feel free to let me know and/or suggest something else. It's just what came to mind for me.

cindyloo:
the Add Custom Collection functionality is broken- the media_ids are not added on the front end, nor on the back end. I'm not sure where it originally fetched them, but we will need to fix this for this and the other related integration issue #1638

dsjen:
Hmm, that seems like an issue! Does this currently work in topic mapper, but not in explorer? If so, is it correct to say that if this PR were to be merged, explorer would remain the same, but topic mapper would lose the ability to use custom collections as sources?

cindyloo:
> Hmm, that seems like an issue! Does this currently work in topic mapper, but not in explorer? If so, is it correct to say that if this PR were to be merged, explorer would remain the same, but topic mapper would lose the ability to use custom collections as sources?

topic mapper actually doesn't use custom collections at this time. It could, but it doesn't at this time

dsjen:
Is it correct to say that explorer won't be able to use custom collection then? I think I'm a little confused b/c it seems like #1638 suggests that explorer can't use them currently? Am I reading that correctly?

dsjen:
Thanks for the info! I really want to make sure that I'm understanding this (as this the functionality is still very new to me) and have specific questions:

* Does explorer currently (i.e. not in this PR) enable use of custom collections?
* Is it correct to say that explorer won't be able to use custom collection then until #1638 is completed?
* If so, why?

cindyloo:
update from tech meeting: make sure the source and collection queries support the keyword and metadata tags, but do not persist the keyword in custom collections

cindyloo:
@rahulbot @dsjen ok I think this is solid. I checked Source Mgr and Explorer if someone wants to test Topic Mapper again. This handles search params into Source Manager, and tags from custom collections will work but a fetch for details of tags is yet to be done. @dsjen we can go over this as part of a learning/skill share

dsjen:
Thanks! It might be worth rebase this off the latest and resolving the conflicts. Also, the PR will need to be made against `main`.

server/views/media_picker.py
```diff
@@ -148,65 +154,74 @@ def _cached_featured_collection_list(tag_id_list):
...
     query = '({})'.format(solr_seed_query)
 
-    if len(media_ids) > 0 or len(tags_ids) > 0 or len(custom_ids):
+    if len(media_ids) > 0 or len(tags_ids) > 0 or len(custom_collection) > 0:
         if tags_ids == [ALL_MEDIA] or tags_ids == ALL_MEDIA:
             return query
         query += " AND ("
         # add in the media sources they specified
-        if len(media_ids) > 0:
+        if len(media_ids) > 0: # this format is a list of media_idds
```
dsjen, line 165: The comment is a little confusing. Shouldn't it be a string of media IDs?

server/views/media_picker.py
```diff
@@ -148,65 +154,74 @@ def _cached_featured_collection_list(tag_id_list):
...
             query += '('+query_media_ids+')'
 
         # conjunction
-        if len(media_ids) > 0 and (len(tags_ids) > 0 or len(custom_ids) > 0):
+        if len(media_ids) > 0 and (len(tags_ids) > 0 or len(custom_collection) > 0):
             query += " OR "
 
         # add in the collections they specified
-        if len(tags_ids) > 0:
+        if len(tags_ids) > 0: # this format is a list of tags_id_medias
```
dsjen, line 177: Same question about the comment

server/views/media_picker.py
```diff
@@ -148,65 +154,74 @@ def _cached_featured_collection_list(tag_id_list):
...
             tags_ids = tags_ids.split(',') if isinstance(tags_ids, str) else tags_ids
             query_tags_ids = " ".join([str(t) for t in tags_ids])
             query_tags_ids = re.sub(r'\[*\]*', '', str(query_tags_ids))
             query_tags_ids = " tags_id_media:({})".format(query_tags_ids)
-            if len(custom_ids) == 0:
+            if not len(custom_collection) > 0:
```
dsjen, line 182: Nit: I'd find this easier to read in the original: `len(custom_ids) == 0`. It took me a second to parse the logic.

server/views/media_picker.py
```diff
@@ -148,65 +154,74 @@ def _cached_featured_collection_list(tag_id_list):
...
-        if len(custom_ids) > 0:
-            query_custom_ids = custom_collection_as_solr_query(custom_ids)
-            if len(tags_ids) > 0:
-                query = "{} OR {} )".format(query, query_custom_ids)  # OR all the sets with the other Collection ids
-            else:
-                query += query_custom_ids  # add the sets to the query (the OR was added before)
+        # grab any custom collections (there may be multiple) and turn it into a boolean tags_id_media phrase
+        if len(custom_collection) > 0: #this format is for each metadata tag, each sub-list is OR'd together, and every list is AND'ed together
+            # we may not have tags_id_media but we may have a keyword
+            #how many?
```
dsjen, line 190: I'm confused by the "how many?" comment

server/views/media_picker.py
```diff
@@ -148,65 +154,74 @@ def _cached_featured_collection_list(tag_id_list):
...
-        return ''
-    custom_ids_dict = json.loads(custom_ids_str)
-    query_custom_ids = ''
-    for sets_of_tags in custom_ids_dict:  # for each custom collections
-        custom_tag_groups = json.loads(sets_of_tags['tags_id_media'])  # expect tags in format [[x, ...], ...]
-        custom_sets = []
-        for tag_grp in custom_tag_groups:
+def custom_collection_as_solr_query(custom_coll_dict):
+    if (len(custom_coll_dict) == 0):
+        return False
```
dsjen, line 204: This method appears to either return a tuple or just a single boolean. Can be be consistent?

Also, it looks like the boolean isn't needed--doesn't seem to be used anywhere. If so, can it be removed?

server/views/media_search.py
```diff
@@ -1,16 +1,37 @@
...
```
dsjen, line 12: Nit: does `request.args.get('linkId', 0)` work?

server/views/media_search.py
```diff
@@ -1,16 +1,37 @@
...
+    else:
+        last_media_id = media_page[len(media_page)-1]['media_id']
+    return media_page, last_media_id
 
 def media_search(search_str, tags_id=None, **kwargs):
     mc = user_mediacloud_client()
     return mc.mediaList(name_like=search_str, tags_id=tags_id, rows=MAX_SOURCES, sort="num_stories", **kwargs)
 
+def collection_search_with_page(search_str, public_only, tag_sets_id_list, **kwargs):
+    link_id = request.args.get('linkId') if 'linkId' in request.args else 0
```
dsjen, line 26: Nit: does `request.args.get('linkId', 0)` work?

src/components/common/mediaPicker/MediaPickerResultsContainer.js
```diff
@@ -26,14 +26,15 @@ class MediaPickerResultsContainer extends React.Component {
     }
   }
 
+  /* unmounting & resetting after query, not what we want
```
dsjen, line 29: Looks like this can just be deleted

src/components/common/mediaPicker/MediaPickerResultsContainer.js
```diff
@@ -161,7 +170,7 @@ const mapDispatchToProps = dispatch => ({
     }
   },
   resetComponents: () => {
-    dispatch(resetMediaPickerQueryArgs());
+    // dispatch(resetMediaPickerQueryArgs());
```
dsjen, line 173: Can be deleted

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
```diff
@@ -0,0 +1,95 @@
...
+import React from 'react';
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { initializePreviouslySelectedMedia, clearSelectedMedia, resetMetadataShortlist } from '../../../actions/systemActions';
+
+function composeMediaPickerSidebarContainer() {
+  return (ChildComponent) => {
+    class MediaPickerSidebarContainer extends React.Component {
+      state = {
+        // queryInStore: false,
```
dsjen, line 11: Can be deleted

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
```diff
@@ -0,0 +1,95 @@
...
+      };
+
+      UNSAFE_componentWillReceiveProps(nextProps) {
```
dsjen, line 14: Can we use something not unsafe here?

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
```diff
@@ -0,0 +1,95 @@
...
+        const { initMedia, handleInitialSelectionOfMedia } = this.props;
+        if (JSON.stringify(initMedia) !== JSON.stringify(nextProps.initMedia)) {
+          if (nextProps.initMedia) { // expects an array of media from caller
+            handleInitialSelectionOfMedia(nextProps.initMedia);
+          }
+        }
+        if ((nextProps.selectedMedia !== this.props.selectedMedia)
+          || (nextProps.selectedMedia && this.props.selectedMedia && nextProps.selectedMedia.length !== this.props.selectedMedia.length)) {
+          // if the results have changed from a keyword entry, we need to update the UI
+          this.shouldComponentUpdate(nextProps);
```
dsjen, line 25: I don't think calling this does anything, does it?

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+      }
+
+      shouldComponentUpdate(nextProps) {
```
dsjen, line 29: This method always returns true. Is that what is intended?

src/components/common/mediaPicker/results/SourceSearchResultsContainer.js
```diff
@@ -270,19 +321,23 @@ const mapStateToProps = state => ({
...
 });
 
 // tags holds metadata search tags
 const mapDispatchToProps = (dispatch, ownProps) => ({
   updateMediaQuerySelection: (values) => {
     if (values && values.tags) {
+      dispatch(resetMediaPickerSources()); // clear prev results
       if (values.allMedia) { // handle the "all media" placeholder selection
-        ownProps.updateMediaQuerySelection({ media_keyword: values.mediaKeyword, type: values.type, allMedia: true });
+        ownProps.updateMediaQuerySelection({ media_keyword: values.mediaKeyword || '*', type: values.type, allMedia: true });
```
dsjen, line 333: `'*'` is referenced a few times in this file and I'm actually not sure what used for. Wildcard? It would be clearer if that were a constant, e.g. `WILDCARD = '*'`

src/components/source/search/MediaPickerSourceSearchContainer.js
```diff
@@ -0,0 +1,59 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Grid } from 'react-flexbox-grid/lib';
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { PICK_SOURCE_AND_COLLECTION } from '../../../lib/explorerUtil';
+import composeMediaPickerSidebarContainer from '../../common/mediaPicker/MediaPickerSidebarContainer';
+// import withAsyncData from '../../common/hocs/AsyncDataContainer';
```
dsjen, line 8: Can be deleted

src/components/source/search/MediaPickerSourceSearchContainer.js
```diff
@@ -0,0 +1,59 @@
...
+import PickedMediaContainer from '../../common/mediaPicker/PickedMediaContainer';
+import MediaPickerResultsContainer from '../../common/mediaPicker/MediaPickerResultsContainer';
+// import { selectMediaPickerQueryArgs, fetchMediaPickerSources } from '../../../actions/systemActions';
```
dsjen, line 11: Can be deleted

src/components/source/search/MediaPickerSourceSearchContainer.js
```diff
@@ -0,0 +1,59 @@
...
+  selectedMedia: PropTypes.array,
+};
+
+const mapStateToProps = state => ({
+  fetchStatus: state.system.mediaPicker.sourceQueryResults.fetchStatus,
+  selectedMediaQueryType: state.system.mediaPicker.selectMediaQuery ? state.system.mediaPicker.selectMediaQuery.args.type : 1,
+  selectedMediaQueryKeyword: state.system.mediaPicker.selectMediaQuery ? state.system.mediaPicker.selectMediaQuery.args.mediaKeyword : null,
+});
+
+/* const fetchAsyncData = (dispatch, ownProps) => {
```
dsjen, line 47: Can be deleted

src/components/source/search/MediaPickerSourceSearchContainer.js
```diff
@@ -0,0 +1,59 @@
...
+  selectedMediaQueryType: PropTypes.number,
+  selectedMediaQueryKeyword: PropTypes.string,
+  onFormChange: PropTypes.func,
+  // from state
+  selectedMedia: PropTypes.array,
+};
+
+const mapStateToProps = state => ({
+  fetchStatus: state.system.mediaPicker.sourceQueryResults.fetchStatus,
+  selectedMediaQueryType: state.system.mediaPicker.selectMediaQuery ? state.system.mediaPicker.selectMediaQuery.args.type : 1,
```
dsjen, line 43: What does it mean for `selectedMediaQueryType` to be `1`?

src/lib/mediaUtil.js
```diff
@@ -1,3 +1,36 @@
+// import { notEmptyString } from './formValidators';
```
dsjen, line 1: Can deleted

src/lib/mediaUtil.js
```diff
@@ -1,3 +1,36 @@
...
+  if (tagsObj) {
+    const updatedTagsObj = { tags: {} };
+    Object.keys(tagsObj).forEach((m) => { // for each tag
+      const vals = Object.values(tagsObj[m]).map(a => a);
+      if (vals && vals.length > 0) {
+        // const tagSet = Object.values(tagsObj[m]).map(a => a).reduce(ts => ts);
+        const readableName = lookupReadableMetadataName(parseInt(m, 10));
+        updatedTagsObj.tags[readableName] = Object.values(tagsObj[m]).map(a => ({ name: readableName, tags_id: a, tag_sets_id: parseInt(m, 10), label: 'Pub Country', tag_set_label: 'Pub Country', tag_set_name: 'pub_country', selected: true, value: true }));
+      }
+      return null;
```
dsjen, line 19: Is this return statement needed? I don't think it does anything?

src/lib/mediaUtil.js
```diff
@@ -1,3 +1,36 @@
...
+import { lookupReadableMetadataName } from './explorerUtil';
+
 export const VIEW_ALL_STORIES = 'VIEW_ALL_STORIES';
 export const VIEW_REGULARLY_COLLECTED = 'VIEW_REGULARLY_COLLECTED';
 export const ALL_MEDIA = -1;
+
+/* this is set up to match the expected params within the metadataCheckboxFieldArray */
+export function importTagsFromCustomSearches(tagsObj) {
+  if (tagsObj) {
+    const updatedTagsObj = { tags: {} };
```
dsjen, line 11: Does this need a `tags` field? It looks like it could be initialized as an empty object and referenced directly instead of using `tags`.

src/lib/mediaUtil.js
```diff
@@ -1,3 +1,36 @@
...
+    Object.keys(tagsObj).forEach((m) => { // for each tag
+      const vals = Object.values(tagsObj[m]).map(a => a);
```
dsjen, line 13: Is the `map` needed here? Seems like it can just be `const vals = Object.values(tagsObj[m])`

src/lib/mediaUtil.js
```diff
@@ -1,3 +1,36 @@
...
+}
+
+export function decodeQueryParamString(queryString) {
+  const queriesFromUrl = queryString; // tags and searchstr
+  if (queryString.query && queryString.query.search) {
+    queriesFromUrl.mediaKeyword = queryString.query.search;
+  }
+  if (queryString.query && queryString.query.tags) {
+    queriesFromUrl.tags = JSON.parse(queryString.query.tags);
+    queriesFromUrl.tags = importTagsFromCustomSearches(queriesFromUrl.tags);
```
dsjen, line 33: I find it confusing for `queriesFromUrl.tags` to be set here and the line above. Can it just rewritten as: `queriesFromUrl.tags = importTagsFromCustomSearches(JSON.parse(queryString.query.tags));`

src/reducers/system/mediaPicker/collectionQueryResults.js
```diff
@@ -1,36 +1,33 @@
 import { FETCH_MEDIAPICKER_COLLECTION_SEARCH, MEDIA_PICKER_TOGGLE_MEDIA_IN_LIST, RESET_MEDIAPICKER_COLLECTION_SEARCH } from '../../../actions/systemActions';
-import { createAsyncReducer } from '../../../lib/reduxHelpers';
+import { createAsyncReducer, concatPrevAndNext } from '../../../lib/reduxHelpers';
 
 const initialState = {
-  args: { type: 0, mediaKeyword: null },
+  args: { type: 1, mediaKeyword: null },
```
dsjen, line 5: What does `1` mean in this context?

dsjen:
I tested this out on Topic Mapper and found a number of issues.

**JS Error in console**
Immediately when I open the media picker in Topic Mapper, I see the following error in the console:

```
EditOpenWebForm.js:32 Uncaught TypeError: Cannot read property 'setState' of undefined
    at setQueryFormChildDialogOpen (EditOpenWebForm.js:32)
    at MediaPickerDialog.eval [as handleModifyClick] (MediaPickerDialog.js:116)
    at onClick (MediaPickerDialog.js:255)
    at HTMLUnknownElement.callCallback (react-dom.development.js:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
    at invokeGuardedCallback (react-dom.development.js:440)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:454)
    at executeDispatch (react-dom.development.js:584)
    at executeDispatchesInOrder (react-dom.development.js:609)
    at executeDispatchesAndRelease (react-dom.development.js:713)
```

**x doesn't go away when no sources selected**
Screen Shot 2020-07-09 at 4 05 36 PM

**No Featured Collections when selecting all media and they don't come back**
Screen Shot 2020-07-09 at 4 09 34 PM

**Tabbing to "Geographic Collection" yields confusing results**
Screen Shot 2020-07-09 at 4 14 44 PM
* The text says there are 20 results (I hadn't searched for anything)
* The text below the results says no results

**No search results for pub state MA**
Screen Shot 2020-07-09 at 4 17 59 PM

**No search results for pub state Georgia and "null"**
Screen Shot 2020-07-09 at 4 18 40 PM

server/views/media_picker.py
```diff
@@ -31,21 +31,25 @@
 @flask_login.login_required
 @api_error_handler
 def api_mediapicker_source_search():
-    search_str = request.args['media_keyword']
+    search_str = request.args['media_keyword'] if 'media_keyword' in request.args else ''
```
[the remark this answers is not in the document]
cindyloo: this pattern is throughout our codebase. Should we change this as we go?

src/lib/mediaUtil.js
cindyloo: true!

src/lib/mediaUtil.js
cindyloo: yes - this was in part my debugging in order to understand the correct URL -> JSON syntax

src/reducers/system/mediaPicker/collectionQueryResults.js
cindyloo: I'll make a constant, but essentially this preserves/determines the tabular setting of collections from sources query settings in the MediaPicker.

src/lib/mediaUtil.js
cindyloo: yes, thank you, it had been setting a property of a

server/views/media_picker.py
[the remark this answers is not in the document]
dsjen: I think so! I find it more readable and a helpful pattern to evolve toward. Thanks for chipping in!

cindyloo:
> I tested this out on Topic Mapper and found a number of issues.
> 
> **JS Error in console**
> Immediately when I open the media picker in Topic Mapper, I see the following error in the console:
> 
> ```
> EditOpenWebForm.js:32 Uncaught TypeError: Cannot read property 'setState' of undefined
>     at setQueryFormChildDialogOpen (EditOpenWebForm.js:32)
>     at MediaPickerDialog.eval [as handleModifyClick] (MediaPickerDialog.js:116)
>     at onClick (MediaPickerDialog.js:255)
>     at HTMLUnknownElement.callCallback (react-dom.development.js:336)
>     at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
>     at invokeGuardedCallback (react-dom.development.js:440)
>     at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:454)
>     at executeDispatch (react-dom.development.js:584)
>     at executeDispatchesInOrder (react-dom.development.js:609)
>     at executeDispatchesAndRelease (react-dom.development.js:713)
> ```
> 
> **x doesn't go away when no sources selected**
> Screen Shot 2020-07-09 at 4 05 36 PM
> 
> **No Featured Collections when selecting all media and they don't come back**
> Screen Shot 2020-07-09 at 4 09 34 PM
> 
> **Tabbing to "Geographic Collection" yields confusing results**
> Screen Shot 2020-07-09 at 4 14 44 PM
> 
> * The text says there are 20 results (I hadn't searched for anything)
> * The text below the results says no results
> 
> **No search results for pub state MA**
> Screen Shot 2020-07-09 at 4 17 59 PM
> 
> **No search results for pub state Georgia and "null"**
> Screen Shot 2020-07-09 at 4 18 40 PM

much of the issues here are related to using the same `collectionResults` in state. The best solution is to reset upon mounting

cindyloo:
collections all now properly resetting as well as properly toggling  esp when no media is then the result. More minor issues can now be tracked #1934 #1935. The bugs fixed above were likely here for a while, so it's good to have hashed them out, thanks @dsjen and @rahulbot

dsjen:
Thanks for the updates! I'm starting to get a little confused about what all there is to do, so am making a list here of what has been previously noted (I haven't reviewed whether they've been addressed):

- [ ] PR should be make against `main`
- [ ] JS `setState` Error in console
- [x] No search results for pub state Georgia and "null" -- Defer to #1935
- [ ] Comments on specific lines in the files are unaddressed
- [ ] Tabbing to "Geographic Collection" yields confusing results
- [ ] No Featured Collections when selecting all media and they don't come back
- [ ] `x` doesn't go away when no sources selected

If any of these have been addressed already, please mark it as done. I'll await your go head to re-review!

cindyloo:
all of these are done with the exception of the PR against main (versus master). I pulled master into the feature brnach and merged..

dsjen:
You're on a roll! Thank you!

I'm confused about the "Comments on specific lines in the files are unaddressed" item though. Is that done? I see many unaddressed comments in the "Files Changed" tab.

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
cindyloo: I don't know of the right way to do this? I'd rather keep the pattern until we modify all the files that use this

src/components/source/search/MediaPickerSourceSearchContainer.js
cindyloo: created a constant, which actually should be shared across the `explorer` and `media` utils files

src/components/common/mediaPicker/results/SourceSearchResultsContainer.js
cindyloo: I think I made a wildcard constant at one point. will find it or make a new one

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
cindyloo: Hooks is new in React 16.8 and I guess useState() would be the answer to this, but I don't think we want to go this route

dsjen: I'm not sure that hooks are the solution here (or maybe there are multiple ones). From https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html:

> getDerivedStateFromProps and getSnapshotBeforeUpdate.

appear to be the replacements.

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
cindyloo: I think we want it to specifically render if those props above are different...I'll take it out to see if it still renders appropriately

server/views/media_picker.py
cindyloo: it is implicitly used. I updated the code to make it explicitly use the boolean, and to return both boolean and query string

src/components/common/mediaPicker/MediaPickerSidebarContainer.js
dsjen: The problem I'm seeing is that it's always returning `true` even when checking the props. Shouldn't one of these return `false`?

server/views/media_picker.py
```diff
@@ -148,66 +154,72 @@ def _cached_featured_collection_list(tag_id_list):
 
 # helper for preview queries
 # tags_id is either a string or a list, which is handled in either case by the len() test. ALL_MEDIA is the exception
-def concatenate_query_for_solr(solr_seed_query, media_ids, tags_ids, custom_ids=None):
-    custom_ids = [] if custom_ids is None else custom_ids
+
+def concatenate_query_for_solr(solr_seed_query, media_ids, tags_ids, custom_collection=None):
```
dsjen, line 158: I'm not sure if you've rebased against the latest, but the value for `custom_collections` should be set to a default array below too. Something like what's done in https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/443a6c8609c2879d51c722ae647f673a6267cfd1/server/views/media_picker.py#L152.

src/components/common/mediaPicker/MediaPickerComponentContainer.js
```diff
@@ -0,0 +1,115 @@
...
+import { connect } from 'react-redux';
+import messages from '../../../resources/messages';
+import PickedMediaContainer from './PickedMediaContainer';
+import MediaPickerResultsContainer from './MediaPickerResultsContainer';
+import { initializePreviouslySelectedMedia, clearSelectedMedia, resetMetadataShortlist } from '../../../actions/systemActions';
+import AppButton from '../AppButton';
+// import { ALL_MEDIA } from '../../../lib/mediaUtil';
+
+class MediaPickerComponentContainer extends React.Component {
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
cindyloo: we have this throughout our codebase. I'd say we should do this all in batch as a separate issue

src/components/common/mediaPicker/results/CollectionSearchResultsContainer.js
```diff
@@ -110,20 +138,42 @@ const mapStateToProps = state => ({
...
   collectionResults: state.system.mediaPicker.collectionQueryResults,
+  links: state.system.mediaPicker.collectionQueryResults.linkId,
 });
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
+  resetAndUpdateMediaQuerySelection: (values) => {
+    if (values && notEmptyString(values.mediaKeyword)) {
+      dispatch(resetMediaPickerCollections());
+      dispatch(selectMediaPickerQueryArgs(values));
+      dispatch(fetchMediaPickerCollections({ media_keyword: (values.mediaKeyword || '*'), which_set: ownProps.whichTagSet || TAG_SET_MC_ID, type: values.type, linkId: values.linkId }));
```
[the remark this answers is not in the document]
cindyloo: the difference is the resetting- we don't want to reset the collections if this is a paging call. The `pageThroughCollections` determines if the `links` property is being passed in, then grabs the value out of the `links.next`. The `search` button calls the method which does reset, the `get more` calls the other.

server/views/media_picker.py
cindyloo: got it. pylinter gave me an 'dangerous' error assigning the custom_collections=[] as default.

cindyloo:
a small handful of minor issues to track #1942 and #1943

src/lib/reduxHelpers.js
```diff
@@ -13,6 +13,19 @@ export function arrayToDict(arr, keyPropertyName) {
...
+export function concatPrevAndNext(next, prev, type) {
+  const prevState = [...prev];
+  const nextList = next.map(c => ({
+    ...c,
+    name: `${c.name}`,
+    id: parseInt(c.media_id, 10),
+    type,
+    selected: false,
+  }));
+  const updatedState = prevState.concat(nextList);
```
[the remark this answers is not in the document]
cindyloo: in our reducers - we are intentionally using "updatedState" to explicitly note the 'updatedState'. We could refactor this in all reducers going forward

src/styles/sources/_search.scss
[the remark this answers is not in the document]
cindyloo: Let's make a new PR to do some of these refactors in the SCSS

cindyloo:
for release notes:

Source Manager
 the Media Picker search functionality replaces the Advanced Search functionality
 removed the ability to create a collection out of source results
 added the ability to page through sources results clicking the "Get Results" button


Explorer, Topic Mapper
fixed metadata tag boolean clauses for multiple custom collections
fixed metadata search tag re-selection issues

src/components/topic/platforms/builder/forms/EditOpenWebForm.js
```diff
@@ -9,63 +10,84 @@ import MediaPickerDialog from '../../../../common/mediaPicker/MediaPickerDialog'
...
+              enableReinitialize
+              keepDirtyOnReinitialize
+              destroyOnUnmount={false}
+              fieldName="media"
+              initialValues={initialValues} // to and from MediaPicker
+              allowRemoval
+              onDelete={onFormChange}
+            />
+            <MediaPickerDialog
+              initMedia={selectedMedia.media ? selectedMedia.media : initialValues.media} // {selected.media ? selected.media : cleanedInitialValues.media}
```
dsjen, line 70: Left over commented out code here?

src/components/source/search/MediaPickerSourceSearchContainer.js
dsjen: I think this was the comment that precipitated creating `MEDIAPICKER_SOURCES_COLLECTIONS_QUERY_SETTING`, but the `1` is still here 😺

src/components/common/mediaPicker/results/SourceSearchResultsContainer.js
dsjen: I think you may have meant to use the newly created `WILDCARD_ASTERISK` here

src/components/common/mediaPicker/results/SourceSearchResultsContainer.js
```diff
@@ -116,62 +127,56 @@ class SourceSearchResultsContainer extends React.Component {
...
         <AppButton
-          style={{ marginTop: -30, float: 'right' }}
+          style={{ float: 'right' }}
           label={formatMessage(localMessages.customColl)}
           onClick={() => this.addCustomSelection({ customColl: true })}
           color="primary"
-          disabled={!selectedMediaQueryTags || Object.keys(selectedMediaQueryTags).length === 0 || sourceResults === undefined || sourceResults.list.length === 0}
+          disabled={!selectedMediaQueryTags
+            || Object.values(selectedMediaQueryTags)
+              // .filter(t => metadataQueryFields.has(t) > 0 selectedMediaQueryTags[t].length > 0)
```
dsjen, line 174: This code can be removed

PR #1879, opened 2020-06-10
Enabled unused-variable linting rule

dsjen:
* Added underscore to unused tuple values 
* Removed unneeded function calls

(Just realized that I made two commits here, but let's just go with it)

* fixed sources CSV upload error
--  Introduced a bug in 234a7c5 😳
-- I'm glad the the linter caught it though 😊

### DIFF
--- a/server/platforms/web_mediacloud.py
+++ b/server/platforms/web_mediacloud.py
@@ -72,7 +72,7 @@ class WebMediaCloudProvider(ContentProvider):
         return top_words
 
     def tags(self, query: str, start_date: dt.datetime, end_date: dt.datetime, **kwargs) -> List[Dict]:
-        q, fq = self._as_query_and_filter_query(query, start_date, end_date, **kwargs)
+        _q, fq = self._as_query_and_filter_query(query, start_date, end_date, **kwargs)
         tags_sets_id = kwargs['tags_sets_id'] if 'tags_sets_id' in kwargs else None
         top_tags = base_apicache.top_tags(query, fq, tags_sets_id)
         return top_tags
--- a/server/views/topics/foci/nyttheme.py
+++ b/server/views/topics/foci/nyttheme.py
@@ -16,7 +16,6 @@ logger = logging.getLogger(__name__)
 
 
 def get_top_themes_by_story_tag_counts(topics_id, num_themes):
-    user_mc_key = user_mediacloud_key()
     nyt_counts = []
 
     #get overall timespan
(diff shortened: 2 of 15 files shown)


server/platforms/web_mediacloud.py
```diff
@@ -72,7 +72,7 @@ def words(self, query: str, start_date: dt.datetime, end_date: dt.datetime, limi
         return top_words
 
     def tags(self, query: str, start_date: dt.datetime, end_date: dt.datetime, **kwargs) -> List[Dict]:
-        q, fq = self._as_query_and_filter_query(query, start_date, end_date, **kwargs)
+        _q, fq = self._as_query_and_filter_query(query, start_date, end_date, **kwargs)
```
rahulbot, line 75: This is an interesting new convention to me. It has always bothered me that those variables are there, but unused. I assume underscoring them means that they are unused in the local scope. I'll adopt that approach. 👍🏽

server/views/topics/foci/nyttheme.py
```diff
@@ -16,7 +16,6 @@
 
 
 def get_top_themes_by_story_tag_counts(topics_id, num_themes):
-    user_mc_key = user_mediacloud_key()
```
rahulbot, line 19: I'm not surprised there are a number of `user_mc_key` unused... those are leftover from when before things moved behind `cache_xxxx` wrappers.

server/views/topics/foci/nyttheme.py
```diff
@@ -58,8 +57,6 @@ def nyt_theme_story_counts(topics_id):
 @arguments_required('numThemes')
 @api_error_handler
 def nyt_theme_coverage(topics_id):
-    # grab the total stories
-    total_stories = topic_story_count(user_mediacloud_key(), topics_id)['count']
```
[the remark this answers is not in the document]
rahulbot: yeah, it is probably leftover from before `topic_tag_coverage` existed

server/views/topics/topicsnapshot.py
```diff
@@ -60,7 +60,7 @@ def _next_snapshot_number(topics_id):
 def topic_snapshot_create(topics_id):
     user_mc = user_mediacloud_client()
     # make a new snapshot
-    new_snapshot = user_mc.topicCreateSnapshot(topics_id, note=_next_snapshot_number(topics_id))['snapshot']
+    user_mc.topicCreateSnapshot(topics_id, note=_next_snapshot_number(topics_id))
```
rahulbot, line 63: Unused return variables like this are probably in place from debugging processes. This is an example where I think the rule is perhaps removing some useful information for developers. I'm OK with that, but wanted to point it out as a tradeoff.

server/platforms/web_mediacloud.py
dsjen: Yes! I find it super helpful too!

server/views/topics/topicsnapshot.py
dsjen: ah, I see! We could also ignore the lint rule on specific lines if it's helpful. I think in general, I do like this rule, but certainly don't think it needs to apply to everything at all times.

PR #1881, opened 2020-06-11
Enabled dangerous-default-value linting rule

dsjen:
It's possible that this fixes some bugs when querying solr.

A nice explanation is at https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python:

> If you modify a default argument, it will persist until the next call, so your "empty" dict will start to contain values on calls other than the first one.


server/views/media_picker.py
```diff
@@ -148,7 +148,9 @@ def _cached_featured_collection_list(tag_id_list):
 
 # helper for preview queries
 # tags_id is either a string or a list, which is handled in either case by the len() test. ALL_MEDIA is the exception
-def concatenate_query_for_solr(solr_seed_query, media_ids, tags_ids, custom_ids=[]):
+def concatenate_query_for_solr(solr_seed_query, media_ids, tags_ids, custom_ids=None):
+    if custom_ids is None:
+        custom_ids = []
```
rahulbot, line 153: It's a nitpick, but I hate wasting lines on silly code like this so I'm wondering if there is a one line solution to this:
* `custom_ids = custom_ids if custom_ids is not None else []`
* or if JS-style syntax work here: `custom_ids = custom_ids or []`

dsjen: I totally agree! ❤️

PR #1882, opened 2020-06-11
Enabled using-constant-test linting test

dsjen:
Just want to pull out this quote:
> The main clause of this if is for when you are editing a media sources, to make sure you skip the current name when checking to see if the media source name is not unique. The else clause is for when you are making a new media source, to make sure it has a unique name.

>I think this mistake was cause the else clause to always evaluate, which means that editing an existing media source probably wasn't (isn't?) working - because it'll hit the else and say that you can't save because it doesn't have a unique name, because it isn't skipping itself when checking for unique names.


server/views/media_picker.py
```diff
@@ -149,7 +149,7 @@ def _cached_featured_collection_list(tag_id_list):
 # helper for preview queries
 # tags_id is either a string or a list, which is handled in either case by the len() test. ALL_MEDIA is the exception
 def concatenate_query_for_solr(solr_seed_query, media_ids, tags_ids, custom_ids=None):
-    custom_ids = [] if None else custom_ids
+    custom_ids = [] if custom_ids is None else custom_ids
```
dsjen, line 152: Ugh, this was my mistake earlier

server/views/sources/search.py
```diff
@@ -50,13 +50,13 @@ def api_collection_search(search_str):
...
     Check if source with name/url exists already
-    :return: boolean indicating if source with this name exists or not (case insensive check)
+    :return: boolean indicating if source with this name exists or not (case insensitive check)
     """
     mc = user_mediacloud_client()
     search_str = request.args['searchStr']
     media_id = int(request.args['id']) if 'id' in request.args else None
     matching_sources = mc.mediaList(name_like=search_str)[:MAX_SOURCES]
-    if id:
+    if media_id:
```
dsjen, line 59: I believe this is what was intended. Might there have been unnoticed bugs w/ `/api/sources/search/name-exists` that this fixes?

rahulbot: The main clause of this `if` is for when you are editing a media sources, to make sure you skip the current name when checking to see if the media source name is not unique. The `else` clause is for when you are making a new media source, to make sure it has a unique name.

I think this mistake was cause the `else` clause to always evaluate, which means that editing an existing  media source probably wasn't (isn't?) working - because it'll hit the else and say that you can't save because it doesn't have a unique name, because it isn't skipping itself when checking for unique names.

That's my guess. So this is a good catch. Linting is proving itself useful pretty quickly 👍🏽

dsjen: It's so amazing! I'm also getting a really tour through the code, as well as refreshing/learning python-y stuff.

PR #1883, opened 2020-06-11
Enabled assignment-from-no-return linting rule

dsjen:


server/platforms/generic_csv.py
```diff
@@ -40,7 +40,7 @@ def _load_from_file(self, path_to_file: str) -> List[Dict]:
         for row in data:
             # this validates the date the same way to back-end parses it
             row['publish_date'] = dateutil.parser.parse(row['publish_date'])
-        self._data = data
+        return data
```
dsjen, line 43: I'm not able to "comment" on it, but check out line 28. I think this possibly fixes a bug where `_data` might be set to `None`.

server/platforms/generic_csv.py
```diff
@@ -16,7 +16,7 @@
 
 class GenericCsvProvider(ContentProvider):
     """
-    User uploads a CSV with content they've gotten from some platofrm we don't provide.
+    User uploads a CSV with content they've gotten from some platform we don't provide.
```
rahulbot, line 19: I'm a terrible speller. Is there a plugin for spell checking in comments in PyCharm?

dsjen: Ha! I think mine was automatically turned on (in preferences > keyword: spelling). No judgement on my end!

dsjen:
I see what you mean about it being subjective. It only caught one case, so certainly doesn't happen often, but I can see how it would prevent misunderstanding.

PR #1886, opened 2020-06-11
Enabled reimported linting rule

dsjen:


dsjen:
Can do! Happy for consistency!

PR #1888, opened 2020-06-11
Enable unsubscriptable-object linting rule

dsjen:


server/views/topics/story.py
```diff
@@ -28,7 +28,7 @@ def story(topics_id, stories_id):
             if tag['tag_sets_id'] == tag_util.GEO_TAG_SET:
                 geonames_id = int(tag['tag'][9:])
                 try:
-                    tag['geoname'] = _cached_geoname(geonames_id, story['language'])
+                    tag['geoname'] = _cached_geoname(geonames_id, story_info['language'])
```
dsjen, line 31: I want to confirm that `story_info` is the correct object to use here. I think the other candidate is `story_topic_info`. I inspected both objects in the debugger and they both have a "language" attribute.

rahulbot: The problem here is that `topicStoryList` and `story`/`storyList` list return different information. So here we (a) query for the `topicStoryList` info for the story specified and (b) query for the regular `story` info - then we merge them. It looks we merge this info into `story_topic_info` and then return that, so in this situation that is the "main" info we should be working with. Because of that I would use `story_topic_info` here, because this happens *after* `story_info` has been merged into `story_topic_info`. I'd also change line 27 to read from `story_topic_info`, so the only place we use `story_info` is to merge its data into `story_topic_info`. Does that make sense?

dsjen: I see! I agree that updating line 27 makes that clearer too.

PR #1889, opened 2020-06-12
Enabled no-member linting rule

dsjen:
* removed unused and broken '/api/explorer/set/<tag_sets_id>' endpoint


server/database.py
```diff
@@ -39,7 +39,7 @@ def add_user(self, username, api_key, profile):
         })
 
     def delete_user(self, username):
-        return self.delete_one({'username': username})
+        return self._conn.users.delete_one({'username': username})
```
rahulbot, line 42: That's a great catch! A few things we need to do now that you found this:
1. This is called in `api_user_delete` (when user deletes themselves). That means the user wasn't actually being deleted from the front-end DB when they deleted themselves from the back-end system. This means we need to queue up a task to audit our front-end  database and remove any user entries that are in there but do not exist in the back end. 
2. I poked around more and found that this is *not* being called in `api_system_user_delete` (admin delete user), even though the should be. We need to add a call to this to remove the user from the front-end DB after they are deleted from the back-end one. The previously mentioned audit also needs to be performed for this reason.

Do those two make sense? Those are the only two places I can think of where we delete users. Can you queue up separate tasks for them?

Why do we have the front-end DB, you might ask? To store things like saved-searches and starred media and collections. Those would have taken longer to build into the back-end and aren't needed there, so we just did it quickly on the front end by making this DB.

rahulbot, line 42: That's a great catch! A few things we need to do now that you found this:

This is called in api_user_delete (when user deletes themselves). That means the user wasn't actually being deleted from the front-end DB when they deleted themselves from the back-end system. This means we need to queue up a task to audit our front-end database and remove any user entries that are in there but do not exist in the back end.
I poked around more and found that this is not being called in api_system_user_delete (admin delete user), even though the should be. We need to add a call to this to remove the user from the front-end DB after they are deleted from the back-end one. The previously mentioned audit also needs to be performed for this reason.
Do those two make sense? Those are the only two places I can think of where we delete users. Can you queue up separate tasks for them?

Why do we have the front-end DB, you might ask? To store things like saved-searches and starred media and collections. Those would have taken longer to build into the back-end and aren't needed there, so we just did it quickly on the front end by making this DB.

server/platforms/test/test_web_google.py
```diff
@@ -4,18 +4,19 @@
...
 
-    def test_fetch_posts_from_api(self):
-        results = self._provider._fetch_posts_from_api(
+    def test_fetch_google_results(self):
+        results = self._provider._fetch_google_results(
             TERM,
             start_date=dt.datetime.now() - dt.timedelta(days=DAY_WINDOW),
-            end_date=dt.datetime.now()
+            end_date=dt.datetime.now(),
+            limit=1
```
rahulbot, line 20: Seems like a better test to fetch more than 1 result here, no?

server/views/explorer/explorer_query.py
```diff
@@ -79,43 +79,3 @@ def api_explorer_searches_by_ids():
         searches_results.append(tag_set_tag_obj)
 
     return jsonify({"results": searches_results})
-
-
-# TODO use this or the other collection list retrieval?
```
rahulbot, line 84: I think you're right and this endpoint isn't used anymore. I searched the `src` dir and couldn't find any references to it.

server/views/explorer/sentences.py
```diff
@@ -20,8 +20,8 @@ def api_explorer_sentences_list():
     if only_queries_reddit(request.form):
         start_date, end_date = parse_query_dates(request.form)
         provider = RedditPushshiftProvider()
-        results = provider.samples(query=request.args['q'], start_date=start_date, end_date=end_date,
-                                   subreddits=NEWS_SUBREDDITS)
+        results = provider.sample(query=request.args['q'], start_date=start_date, end_date=end_date,
+                                  subreddits=NEWS_SUBREDDITS)
```
rahulbot, line 24: This is related to a temporary hack I put in place to allow us to search Reddit submissions in Explorer - if the magical reddit news source is selected as the only source. It is probably worth ripping that out as part of code simplification. You can preview Reddit results in the topic platforms workflow, and building a generic platform Explorer isn't within scope.

server/database.py
dsjen: Ah, I see! I'll create those tasks.

server/platforms/test/test_web_google.py
dsjen: It all seems little arbitrary to me and the point is that the thing returns a list. I'll set it to 10 just to get more results?

server/views/explorer/explorer_query.py
dsjen: Excellent! I love deleting code!

server/views/explorer/sentences.py
dsjen: I see. Shall I queue up a task for removing that functionality then?

rahulbot: How about we double check on Tues with the research folks to see if anyone is using it? I would guess not... it was kind of a stop-gap to "get to know" PS reddit data.

dsjen: I see! Sounds like a plan.

PR #1892, opened 2020-06-12
Enable useless-import-alias linting rule

dsjen:


dsjen:
That make sense to me! I'll add that to the coding conventions

PR #1894, opened 2020-06-12
Enabled superfluous-parens linting rule

dsjen:


dsjen:
You know, I was too! I was almost going to ignore this rule, but when I ran it saw that it was just once case. I figure that we might as well codify this, since we're following it already.

PR #1898, opened 2020-06-12
Enabled unused-argument linting rule

dsjen:
The linter pointed out a bunch of places where it seemed arguments were no longer used. I removed the ones that didn't seem needed (using the refactoring tooling in pycharm). Some were seemingly unused, but actually used as keys for caching or used as part of the routes, so I added an underscore to those.

There were a handful of places that it seemed like caching was desired, but it wasn't enabled, so I added the caching decorator to those places.


server/views/sources/apicache.py
```diff
@@ -78,8 +78,9 @@ def _cached_timeperiod_story_count(q='*', time_period=QUERY_LAST_MONTH):
 MC_START_DATE = datetime.datetime(2010, 1, 1, 0, 0, 0)  # kind of when media cloud started
 
 
-# you can specify last_n_days to be 365 if you only want the last year of results
-def split_story_count(user_mc_key, q='*', last_n_days=None):
+@cache.cache_on_arguments()
+def split_story_count(_user_mc_key, q='*', last_n_days=None):
```
rahulbot, line 82: I think one doesn't need the `cache` decorator. In general we've cached as close to the API call as possible (because that is the only slow thing in comparison to any other compute-bound work). In this case, `_cached_split_story_counts` is being cached, and this method does very little computational work it, so if I've read it correct there is no need to cache it. Accordingly, I think the `_user_mc_key` argument could be removed.

server/views/sources/geocount.py
```diff
@@ -11,10 +11,10 @@
 
 
 @cache.cache_on_arguments()
-def cached_geotag_count(user_mc_key, query):
+def cached_geotag_count(_user_mc_key, query):
```
rahulbot, line 14: I think you can ditch the `_user_mc_key` here. It is fine to have the geotag counts for a source be a cross-user cache. There aren't any user-based permissions on this data.

server/views/topics/apicache.py
```diff
@@ -357,13 +358,13 @@ def topic_tag_counts(user_mc_key, topics_id, tag_sets_id):
...
 @cache.cache_on_arguments()
-def _cached_topic_tag_counts(user_mc_key, topics_id, tag_sets_id, query):
+def _cached_topic_tag_counts(_user_mc_key, _topics_id, tag_sets_id, query):
     # even though we call base_apicache under the hood here, we want to make sure the cache is keyed by
     # API key, because topics have user-level permissioning
     return base_apicache.top_tags(query, None, tag_sets_id=tag_sets_id)
 
 
-def topic_sentence_sample(user_mc_key, topics_id, sample_size=1000, **kwargs):
+def topic_sentence_sample(user_mc_key, sample_size=1000, **kwargs):
```
rahulbot, line 367: Perhaps worth adding a comment here about why `topics_id` isn't needed... it is because there is no `topicSentenceList` endpoint. So to pull random sentence samples we by the `timespans_id` with a regular  `sentenceList` call.

server/views/topics/apicache.py
```diff
@@ -320,7 +320,8 @@ def topic_focal_set(user_mc_key, topics_id, snapshots_id, focal_sets_id):
     raise ValueError("Unknown subtopic set id of {}".format(focal_sets_id))
 
 
-def cached_topic_timespan_list(user_mc_key, topics_id, snapshots_id=None, foci_id=None):
+@cache.cache_on_arguments()
+def cached_topic_timespan_list(_user_mc_key, topics_id, snapshots_id=None, foci_id=None):
```
rahulbot, line 324: I have a recollection of @cindyloo saying we shouldn't cache. I think the assumption I had was that the list of timespans within a snapshot don't ever change. However, I think Cindy pointed out that in our new multi-platform world they do, because a snapshot is empty for a bit while you are adding platforms. Once you generate the snapshot the filters won't ever change, but before then this call will be empty.

This should probably be renamed and refactored *out* of this `apicache` module.

server/views/sources/apicache.py
dsjen: Gotcha!

dsjen: Actually, in looking at this again, I think I'm a little confused on the caching bit. `split_story_count()` is used in two places:

https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/234a7c57b65c671c4828df8e39cbdc417d6f915a/server/views/sources/collection.py#L261

https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/8272c855349b82febaf4fbbaa9d25c82a801729c/server/views/sources/stories_split_by_time.py#L13
* Upon further inspection `stream_split_stories_csv()` doesn't seem used anywhere, and **I propose deleting it**. Is there a reason it should be kept around?

Also, just to be clear, `_cached_split_story_counts()` doesn't call `split_story_count()`. It ends up being called by the `/api/collections/<collection_id>/sources/story-split/historical-counts` endpoints.

server/views/topics/apicache.py
dsjen: Ok, how about I remove the caching and use of `user_mc_key` in this PR then and create a new issue to refactor this logic out. It's used in a bunch of difference places, and I'd rather isolate that change in a separate commit.

server/views/sources/apicache.py
dsjen: Deleted `stream_split_stories_csv`

server/views/topics/apicache.py
rahulbot: Makes sense to me.

dsjen: Issue: #1905

PR #1899, opened 2020-06-12
Enable too-many-format-args linting rule

dsjen:
The `editor_notes` probably looked a little odd. I've adjusted the sentence so that all the arguments are incorporated now.


dsjen:
Pretty neat, huh? I found myself confused, so figured might as well as the keywords.

PR #1900, opened 2020-06-12
Enabled redefined-outer-name linting rule

dsjen:


dsjen:
Ha! I keep telling myself, "just one more..."

dsjen:
Ah! Yes, so the diff doesn't show it, but the issue was that `story` and `media` were both methods (in addition to being variables). One might inadvertently try to use the method/variable (as well as not being able to call the method at all) without renaming the variables. Alternatively, I could have renamed the methods I suppose.

PR #1901, opened 2020-06-12
Enabled no-value-for-parameter linting rule

dsjen:


server/views/topics/provider.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger(__name__)
 
 
-def _parse_optional_args(arg_names: List, custom_defaults: Dict) -> Dict:
+def _parse_optional_args(arg_names: List, custom_defaults: Dict = None) -> Dict:
```
dsjen, line 20: a few methods were calling this w/o custom_default (I think I forgot to add the default value back in another refactor)

PR #1902, opened 2020-06-15
Enabled anomalous-backslash-in-string linting rule

dsjen:


server/util/tags.py
```diff
@@ -121,7 +121,7 @@ def is_bad_theme(tag_id):
 
 
 def format_name_from_label(user_label):
-    formatted_name = re.sub('\W|^(?=\d)', '_', user_label)
+    formatted_name = re.sub(r'\W|^(?=\d)', '_', user_label)
```
rahulbot, line 124: This seems to be called when editing collections to automatically generate the `tag` string from the human-readable name the user inputted. I am reading what it does correctly? If so I have one thought: move this function to `stringutil` and name it something like `as_tag_name`. Not necessarily, but might be more understandable in the future.

dsjen: That makes sense! I'm on it.

PR #1906, opened 2020-06-16
Leave remaining linting rules disabled

dsjen:
I propose leaving the remaining rules disabled. Please let me know what you think!


dsjen:
Yes, that was my sense too. We can chip away at it as time permits or the desire bubbles over. It's actually pretty informative to see what's flag. None of the ones that I left in seemed to catch any bugs when I enabled them FWIW.

Issue #1907, opened 2020-06-17
Explorer missing bigram / trigram downloads


rahulbot:
Reported by a user. I must have removed them accidentally as a by-product of my refactoring work.

rahulbot:
For some reason, the bigram and trigram `MenuItem`s in `EditableWordCloudDataCard` were coded to only show up if there was an `onDownload` handler defined. This seems to be encoding some implicit logic about when to show them, so I poked around the usage of `onDownload` in parent Components but couldn't find any logic to this. Maybe previously other places didn't support bigram/trigram download? 🤷🏽‍♂️ So I've left them enabled no matter if `onDownload` or `downloadURL` is defined.

PR #1908, opened 2020-06-17
fix explorer ngram download problem

rahulbot:
Details on #1907

### DIFF
--- a/src/components/common/EditableWordCloudDataCard.js
+++ b/src/components/common/EditableWordCloudDataCard.js
@@ -123,7 +123,7 @@ class EditableWordCloudDataCard extends React.Component {
   };
 
   buildActionMenu = (uniqueDomId) => {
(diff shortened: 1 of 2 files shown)


src/components/common/EditableWordCloudDataCard.js
```diff
@@ -207,27 +207,23 @@ class EditableWordCloudDataCard extends React.Component {
...
+        <ListItemIcon>
+          <DownloadButton />
+        </ListItemIcon>
+      </MenuItem>,
+      <MenuItem
+        className="action-icon-menu-item"
+        disabled={this.state.editing} // can't download until done editing
+        onClick={() => this.downloadCsv(3)}
+      >
+        <FormattedMessage {...localMessages.downloadTrigramCSV} />
```
dsjen, line 225: Do we want a download icon here too?

Screen Shot 2020-06-17 at 2 27 30 PM

rahulbot: Yes - good catch. It should be consistent.

PR #1918, opened 2020-06-18
Use "version" from package.json

dsjen:
I think we agreed to just source the version from once place. I think `package.json` would be a natural place to store this.

### DIFF
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -22,7 +22,7 @@ const baseConfig = {
     new MomentLocalesPlugin(),
     // provide much cleaner feedback while building on the command line
     new WebpackBar(),
-    new webpack.DefinePlugin({ MC_VERSION: JSON.stringify(require('../package.json').version) }),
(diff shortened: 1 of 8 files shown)


config/webpack.base.config.js
```diff
@@ -22,6 +22,7 @@ const baseConfig = {
     new MomentLocalesPlugin(),
     // provide much cleaner feedback while building on the command line
     new WebpackBar(),
+    new webpack.DefinePlugin({ MC_VERSION: JSON.stringify(require('../package.json').version) }),
```
rahulbot, line 25: This uses a relative path, but it looks like all the other paths in here use `path.resolve(basedir, 'some', 'set', 'of', 'dirs')`. It would make sense to harmonize these to one standard. I don't have a strong reason for requiring absolute paths, except that it works and I've found web pack stuff to be *incredibly* brittle, so I'm inclined to suggest changing this line to match the others (ie. be an absolute path).

dsjen: Yes, let's use the `path.resolve`! Like you say, it's consistent, and I'd rather go with whatever is working, especially in webpack land!

PR #1922, opened 2020-06-24
Delete user from frontend DB when an admin deletes user

### LINKED ISSUE #1891: Delete user in api_system_user_delete
Related to #1889, we discovered that `database.delete_user()` isn't called in `api_system_user_delete`, and it should be. In this task, `delete_user()` should be called there.

dsjen:
Fixes #1891


dsjen:
I did test it out locally too :)

Issue #1890, opened 2020-06-12
Clean up users on frontend


dsjen:
In #1889, we discovered a bug where user data wasn't cleared out from the frontend DB when users remove their accounts. In this task, we need to audit our frontend DB and remove any entries not found in the backend DB.

Tasks:

- [x] Write script for deleting
- [ ] Run script

rahulbot:
Relevant example in our code searching for a user by email; https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/16018aeb5228ed958c0c80419cbf38425da93e83/server/views/admin/users.py#L18

That's part of the code under the Admin->Manage Users menu.

Issue #1727, opened 2020-01-23
switch to python-dotenv for config


rahulbot:
Right now we use a custom config class I created years ago, before I knew [python-dotenv](https://pypi.org/project/python-dotenv/) existed. We should switch to using python-dotenv, because it is better and a standard thing people use to load config vars from a file (.env) locally, but from env vars in a container deployment. This will involve work refactoring to remove any uses of the config object, to instead read those items out of env vars directly. This could perhaps also involve centralizing the API-Key driven services into factory methods that return the objects, so all of their instantiation is in one place. Happy to discuss approaches we all like further.

PR #1923, opened 2020-06-25
Added sync-frontend-db custom command

dsjen:
First part of #1890

* Adds `flask sync-frontend-db`
* Adds `..flaskenv` for convenience in running flask command line

The new command calls the backend API to fetch users, diffs with the frontend, and removes users extra users.

N.B. PR will merge to `3.14.x`.

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -12,6 +12,7 @@ mediacloud-cliff==2.6.1
 gunicorn==20.0.4
 dogpile.cache==0.7.1  # upgrading dogpile.cache caused problems - leave it on this version
 python-dateutil==2.8.1
+python-dotenv==0.13.0
 python-slugify==4.0.0
 greenlet==0.4.15
 deco==0.5.2
--- a/server/commands/__init__.py
+++ b/server/commands/__init__.py
@@ -4,7 +4,7 @@ import click
 from flask.cli import with_appcontext
 
 
-def fetch_backend_emails():
+def _fetch_backend_emails():
     from server import mc
     first = True
     link_id = None
@@ -27,7 +27,7 @@ def sync_frontend_db(test):
     will be deleted.
     """
     from server import user_db
-    backend_emails = fetch_backend_emails()
+    backend_emails = _fetch_backend_emails()
     print("'{}' backend users".format(len(backend_emails)))
 
     users_to_remove = user_db.get_users({
--- a/server/commands/__init__.py
+++ b/server/commands/__init__.py
@@ -1,0 +1,48 @@
+# pylint: disable=import-outside-toplevel
+
+import click
+from flask.cli import with_appcontext
+
+
+def fetch_backend_emails():
+    from server import mc
+    first = True
+    link_id = None
+    backend_emails = []
+    while first or link_id:
+        first = False
+        page = mc.userList(link_id=link_id)
+        link_id = page["link_ids"].get("next")
+        backend_emails = backend_emails + [user["email"] for user in page["users"]]
+    return backend_emails
+
+
+@click.command("sync-frontend-db")
+@click.option('--test', required=False, type=bool, default=False,
+              help="Fetches and display number of users to delete only. Doesn't perform delete.")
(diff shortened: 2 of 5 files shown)


requirements/common.txt
```diff
@@ -12,6 +12,7 @@ mediacloud-cliff==2.6.1
 gunicorn==20.0.4
 dogpile.cache==0.7.1  # upgrading dogpile.cache caused problems - leave it on this version
 python-dateutil==2.8.1
+python-dotenv==0.13.0
```
rahulbot, line 15: This is good prep for #1727

server/database.py
```diff
@@ -66,6 +66,12 @@ def remove_item_from_users_list(self, username, list_name, item):
     def update_user(self, username, values_to_update):
         return self._conn.users.update_one({'username': username}, {'$set': values_to_update})
 
+    def get_users(self, query_operator=None):
+        return self._conn.users.find(query_operator or {})
+
+    def delete_users(self, query_operator=None):
+        return self._conn.users.delete_many(query_operator or {})
```
rahulbot, line 73: Just to confirm, calling `delete_many({})` matches none, not all, right? If not then this is a bit dangerous and I'd say `query_operator` shouldn't default to `None` here.

server/commands/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        page = mc.userList(link_id=link_id)
+        link_id = page["link_ids"].get("next")
+        backend_emails = backend_emails + [user["email"] for user in page["users"]]
+    return backend_emails
+
+
+@click.command("sync-frontend-db")
+@click.option('--test', required=False, type=bool, default=False,
+              help="Fetches and display number of users to delete only. Doesn't perform delete.")
+@with_appcontext
```
rahulbot, line 23: This is a great approach for running cli commands within the app context 👍🏽 I didn't know about it.

server/commands/__init__.py
```diff
@@ -0,0 +1,48 @@
+# pylint: disable=import-outside-toplevel
+
+import click
+from flask.cli import with_appcontext
+
+
+def fetch_backend_emails():
```
rahulbot, line 7: Should this be private? ie. `_fetch_backend_emails()`

server/commands/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+@click.option('--test', required=False, type=bool, default=False,
+              help="Fetches and display number of users to delete only. Doesn't perform delete.")
+@with_appcontext
+def sync_frontend_db(test):
+    """
+    Synchronize the frontend user database with the backend. Users in the frontend database not found in the backend
+    will be deleted.
+    """
+    from server import user_db
+    backend_emails = fetch_backend_emails()
```
rahulbot, line 30: How many users do we have right now - does this take a long time?

server/database.py
dsjen: That's so true! I'll update this.

server/commands/__init__.py
dsjen: This takes a very long time! Let me time it just so we have an idea. I'll report back in a bit.

server/commands/__init__.py
dsjen: Isn't it great?!?

requirements/common.txt
dsjen: I hadn't noticed that ticket. We're on the same page!

server/commands/__init__.py
dsjen: Actually, it seemed like it was _forever_ initially, but it's 5 minutes, and we have a 10.5K users.

dsjen:
I'll plan on deploying early next week and running the script.

server/commands/__init__.py
rahulbot: That seems ok for this one-time operation. The alternative is listing users in our front end DB and making one query checking if each still exists in the backend dB. Either seems ok to me.

PR #1947, opened 2020-07-22
Fixed compare top words

### LINKED ISSUE #1924: Comparative word clouds missing from Explorer
It looks like the comparative word clouds are missing from Explorer. When I scroll from the word space model to the themes, there's text saying "error" off to the side and no comparative word clouds. (Regular word clouds at the top of the 'language' tab are working.) 

Screen Shot 2020-06-26 at 11 28 59 AM

dsjen:
Fixes #1924 

This reinstates the word cloud comparisons and updates the error message. It looks like the error stemmed from a change in the `results`. It appears to be an dict and all the code was expecting an array. I also removed the `OrderedWordCloud` in the case that there isn't a `rightQuery`. I don't see how this is possible and suspect that this is old code.

**2 Queries**
Screen Shot 2020-07-22 at 3 48 45 PM

**More than two queries**
Screen Shot 2020-07-22 at 3 50 05 PM

**Error message**
Screen Shot 2020-07-22 at 4 06 54 PM


dsjen:
No worries! I thought it was really neat to see back in action.

PR #1949, opened 2020-07-23
Replaced 2016 with 2020 US Partisanship Sets

### LINKED ISSUE #1931: Update featured partisan sources in collections search tool
The front end marks only certain tags to be returned as collections in the source/collection search tool. Currently `tag_sets_id = 1959` is the partisan sets. We should remove those tags and instead return the tags from `tag_sets_id = 15765109`

The new partisan tags should be added to the 'featured collections' screen in the collection search tool.

TODOs:
- [ ] Added 2020 tag set to featured list
- [ ] Rename 2016 tag set to include "2016" on them

dsjen:
Fixes #1931


dsjen:
Thanks! When this is deployed, I'll definitely try to clear the redis cache to make sure my permissions work!

PR #1962, opened 2020-08-06
Updated featured pub state collections, fixed search title

### LINKED ISSUE #1935: MediaPicker may not show Pub State results correctly
queries on Pub State like "massachussets" or "georgia" result in 0 results.

also, below is a a typo that sometimes shows up

Screen Shot 2020-07-10 at 7 58 48 PM

dsjen:
Fixes #1935

I've updated the Massachusetts, California, and Georgia collections to point to the state and local version of each. The previous ones contained no sources.

The other changes I made was to fix the `null` that was appearing in the media picker search results. This happens when searching with a new tag (after the first one). I'll mark in the file where the key change was.

**New**
Screen Shot 2020-08-06 at 12 54 47 PM

**Old**
Screen Shot 2020-08-06 at 12 55 50 PM


src/components/common/mediaPicker/results/SourceSearchResultsContainer.js
```diff
@@ -180,36 +175,42 @@ class SourceSearchResultsContainer extends React.Component {
...
           if (sourceResults.args.tags) { // correlate searched tag ids with objects so we can display the labels
-            if (!previouslySearchedTags[i]) previouslySearchedTags[i] = [];
-            previouslySearchedTags[i] = obj.map(t => ( // if in tags, it is selected, so reflect this
-              sourceResults.args.tags.indexOf(t.tags_id) > -1 ? ({ ...t, selected: true }) : ''
-            ));
+            if (!previouslySearchedTags[i]) {
+              previouslySearchedTags[i] = [];
+            }
+            previouslySearchedTags[i] = tag
+              .filter(t => sourceResults.args.tags.indexOf(t.tags_id) > -1)
```
dsjen, line 193: Filtering the results was key to fixing the `null`. Previously, this was inserting an empty string, which was converted into `null` in `explorerUtils::stringifyTags()`.

cindyloo: nice and thanks! it's hairy in there

dsjen: So many things to consider!

PR #1968, opened 2020-08-11
Remove all static tags and switch to discovery-based process

### LINKED ISSUE #1961: Build generic solution for required tags and tag-sets
We have a number of tags and tag-sets that are hard-coded into our system, and are needed to make the UI work. We should:

- [x]  list these tags and tag-sets that need to be hard-coded
- [x]  figure out which are required
- [x] come up with some generic solution that centralizes them and allows them to be customized for a different install (which will have different `tags_id` and `tag_sets_id` values)

Related to #1960.

rahulbot:
We have static tag IDs scattered throughout our code as constants. However, these only work on our primary install of Media Cloud. To remedy this, I've created a discovery-based approach that relies on well-know tag names. Anyone installing media cloud will need to make sure those tag / tag-set names are in their database. Based on those names, the relevant `tags_id` and `tag_sets_id` values will be discovered by the server (in `tags.py`) and relayed over to be used in the front end (in `state.system.staticTags`.

This code isn't complicated, but touched a lot of different parts of our system. I've clicked around, but it is very likely I got a few of these spots adapted incorrectly because there were just so many of them. Please review my overall approach, but also checkout and click around each tool or spots you think are particularly brittle.

I'll continue to commit fixes as I find them. This fixes #1961.


dsjen:
This is really interesting and helps to simplify the codebase! I'm still looking through it and will get it running soon, but I wanted to let you know I've got eyes on this PR. Sorry it's been languishing here!

dsjen:
How about disabling it just for the file by adding disable comment, e.g. https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/443a6c8609c2879d51c722ae647f673a6267cfd1/server/platforms/test/test_web_google.py#L1

Thanks!

dsjen:
FYI, I'm getting an error when clicking on the "versions" tab in topic mapper on this branch:

```
  File "/Users/dsjen/development/MediaCloud-Web-Tools/server/views/topics/stories.py", line 234, in story_counts_by_snapshot
    tag_util.TagDiscoverer.is_spidered_story_tag))['count']
AttributeError: type object 'TagDiscoverer' has no attribute 'is_spidered_story_tag'
```

dsjen:
I tested out the media picker in Explorer and see a problem with search. In the "Geographic Collections" tab, when I search for "united states," I get 0 results. When switching back to the main branch I see 20 results.

dsjen:
Thank you for fixing that! Sorry for the slow trickle, but I found another one on the "Suggest A Source" page. In the last field, "Search for collections", no results are returned. I tried this out on this branch (nothing returns results) and the `main` branch (get results).

dsjen:
Thank you! I clicked around on all apps and found one error on source manager at `#/collections/media-cloud`

```
Warning: Failed prop type: The prop `collectionSets` is marked as required in `MCCollectionListContainer`, but its value is `undefined`.
```

I'm so sorry for another issue trickling into this and hope that it's not too annoying. I hope with my extensive clicking around that this is the last bug. Thank you!

esirK:
Thank you for this it solves Issues I'm having with my deployment. Is it safe to say it's ready for merging? If not, could you share with me the tags I would need to create manually? e.g I've noticed requests to ` /api/v2/tag_sets/single/1935`

dsjen:
Yes, please do!

PR #2007, opened 2020-11-13
Allow use of different mediacloud api backends

esirK:
Adds the ability to use different mediacloud API backends by specifying a **MEDIA_CLOUD_API_URL** environment variable.
Fixes https://github.com/mediacloud/web-tools/issues/1958

### DIFF
--- a/server/util/tags.py
+++ b/server/util/tags.py
@@ -8,6 +8,7 @@ import codecs
 from server import base_dir
 from server.auth import user_mediacloud_client
 from server.cache import cache
+from server.util.config import get_default_config
 
 logger = logging.getLogger(__name__)
 
@@ -69,6 +70,9 @@ VALID_METADATA_IDS = [
 ]
 
 TAG_SPIDERED_STORY = 8875452
+
+# load the config helper
+config = get_default_config()
 
 
 def processed_for_themes_query_clause():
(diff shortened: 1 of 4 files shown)


server/util/tags.py
```diff
@@ -144,6 +148,7 @@ def tags_in_tag_set(mc_api_key, tag_sets_id):
 def tag_set_with_tags(mc_api_key, tag_sets_id, only_public_tags=False, use_file_cache=False):
     # don't need to cache here, because either you are reading from a file, or each page is cached
     local_mc = MediaCloud(mc_api_key)
+    local_mc.V2_API_URL = config.get('MEDIA_CLOUD_API_URL')
```
rahulbot, line 151: This seems like a pre-existing bug - shouldn't the line above instead call this from auth.py?
```python
local_mc = user_mediacloud_client(mc_api_key)
```
Then the extra line here wouldn't be needed and the code would be centralized

rahulbot (changes requested):
This looks great. Thanks! 

One suggestion: It'd be nice if this was an optional config item, so we don't need to set it on all our production servers. It looks like `config.get` throws a `KeyError` if a env var isn't defined, so this would be a little better for us in each place:
```python
try:
    mc.V2_API_URL = config.get('MEDIA_CLOUD_API_URL')
except KeyError:
    pass # just use the default API url because a custom one is not defined
```
What do you think @dsjen?

server/util/tags.py
esirK: Sure. This makes sense.

dsjen (approved):
Thank you so much! This looks great!

PR #2047, opened 2021-02-08
Remove email config

dsjen:
- [ ]  Remove settings on containers


dsjen:
Turns out mail gets send w/ batch updates to collections

PR #2049, opened 2021-02-10
update email template submod

jtotoole:
See https://github.com/mediacloud/systems/issues/218, https://github.com/mediacloud/email-templates/pull/2


dsjen (approved):
LGTM! Thanks!

PR #2068, opened 2021-03-29
fix country iso3166 info in static country collection list

rahulbot:
The static list of country collections didn't have the right iso3166 info (alpha2, numeric, etc). This was breaking a notebook script I was running to make a list of how many stories we have from each counties last year ([dataviz!](https://observablehq.com/d/54808ebde52058e1?ui=classic)). I changed this script to generate that static json with the correct metadata.

I verified that the Source Mgr "view geographic collections" page still loads (`'/api/collection/set/geo-by-country). I can't find anywhere else this file is used.


dsjen:
Thank you! Something has gone awry with my system (not related of this PR), and I'll review this after I get this resolved...

dsjen:
I tested this out and everything functions. This change also creates separate categories for the following countries:
* Papua New Guinea
* South Sudan
* Taiwan, Province of China

Is that intentional? What was the reasoning for not separating them before?

dsjen (approved):

dsjen:
Just noticed this is still here. I think it's good to go. Feel free to merge!

PR #2071, opened 2021-04-10
Change package scripts for cross system environment variables

amifunny:
Declaring environment variable like this `MC_APP_NAME=explorer` caused `MC_APP_NAME` not found error in windows. To solve this, scripts are prefixed with `cross-env` which is [package](https://www.npmjs.com/package/cross-env) just for this. It works for any operating system so it won't affect Linux development environment and just extend the same code to be run by windows without manually changing it after cloning.

In `package.json`, dependencies are also edited to add `cross-env`. Also edited `.eslint` to ignore line breaking which was causing the error `Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style` for windows system.


package.json
```diff
@@ -27,6 +27,7 @@
     "codemirror": "^5.54.0",
     "core-js": "^3.6.5",
     "country-iso-3-to-2": "^1.1.0",
+    "cross-env": "^7.0.3",
```
dsjen, line 30: Please move this to the `devDependencies` section

dsjen (changes requested):
Thank you so much for the fix! I tested this out and it works. Please move the package to the `devDependencies` section.

package.json
amifunny: Yes, of course it should should be in dev. I have made the change and moved it to `devDependencies`.

dsjen (approved):
This is great--thank you!

amifunny:
@dsjen thank you for reviewing the PR

